PR #1, opened 2020-12-10
done

nohaderf:
:)

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,10 +1,14 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+let counter = 1
+counter = 2
+
+console.log("%cQuestion 1", "color: red") 
+
+console.log(counter) 
+// => 2
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,10 +1,14 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+let counter = 1
+counter = 2
```
BrendaMichelle, line 5: Nice 💯  `let` allows us to reassign our variables

index.js
```diff
@@ -13,76 +17,117 @@
 // *** otherwise, the error will stop the rest of your code from running
 // console.log("%cQuestion 2", "color: red") 
 
-// console.log(name)
+const name = "Freda"
```
BrendaMichelle, line 20: Great! The key here was using `const` because it doesn't allow us to reassign our variable

index.js
```diff
@@ -13,76 +17,117 @@
...
+//   current_thirst_level -= 1
+//   puts "Ahh that hits the spot"
+//   current_thirst_level
+// end
+
+function drinkWater(currentThirstLevel){
+  console.log("Man I sure am thirsty")
+  currentThirstLevel -= 1
+  console.log("Ahh that hits the spot")
+  return currentThirstLevel
```
BrendaMichelle, line 45: Looking good

index.js
```diff
@@ -13,76 +17,117 @@
...
+function sameSameButDifferent(num, maybeNum){
+  if (num === maybeNum){
+    return "same same"
+  } else if (num == maybeNum){
+    return "same same (but different)"
+  } else {
+    return "different"
+  }
+}
+
```
BrendaMichelle, line 68: Nice! The key here was understanding the difference between `===` and `==`. You can also convert the string to a number with parseInt to avoid == :

index.js
```diff
@@ -13,76 +17,117 @@
...
-// console.log("%cQuestion 5", "color: red")
-// const student1 = { name: "Duane", grade: 88 }
-// updateGrade(student1, 92)
-// console.log(student1)
-// // => { name: "Duane", grade: 92 }
-// console.log("%c----------", "color: red")
+
+function updateGrade(student, grade){
+  student.grade = grade
+}
```
BrendaMichelle, line 91: Nice! The key here was understanding that objects are reference types. Object values are references, so when we mutate the `student` object inside `updateGrade`, the change persists outside of the function.

index.js
```diff
@@ -13,76 +17,117 @@
...
-// // => "Duane"
+
+function printNameAndPhones(users){
+  users.forEach(function(user){
+    console.log(user.name)
+    console.log(`Cell: ${user.phones.cell}`)
+    console.log(`Office: ${user.phones.office}`)
+  })
+}
+
```
BrendaMichelle, line 112: Nice. callback function as an argument to `forEach`  ✅  template strings for string interpolation ✅
Btw, nice work naming the anonymous function's parameter something semantic!

index.js
```diff
@@ -99,24 +144,31 @@ function myMap(array, callback) {
 
 // ***** Callbacks - Question 1 *****
 
+function triple(num){
+  return num * 3
+}
```
BrendaMichelle, line 149: Looks good

index.js
```diff
@@ -99,24 +144,31 @@ function myMap(array, callback) {
...
 // ***** Callbacks - Question 2 *****
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 2", "color: red")
 
-// console.log(myMap(["Raffy", "Chase"], greet)) 
-// // => ["Hi, Raffy!", "Hi, Chase!"]
-// console.log("%c----------", "color: red") 
+function greet(name){
+  return(`Hi, ${name}!`)
+}
```
BrendaMichelle, line 166: Looks good, too

index.js
```diff
@@ -141,44 +193,63 @@ const line = []
...
-// console.log("%cScope & Closures - Question 1", "color: red")
 
-// console.log(takeATicketNumber(line))
-// // => `Welcome. You are ticket number 1`
+let ticket = 0
+function takeATicketNumber(line){
+  ticket ++
+  line.push(ticket)
+  return (`Welcome. You are ticket number ${ticket}`)
+}
```
BrendaMichelle, line 202: awesome! They key here was understanding we needed to use a global variable

index.js
```diff
@@ -141,44 +193,63 @@ const line = []
...
-// // => `Welcome. You are ticket number 1`
+function ticketNumberGeneratorFunc(){
+  let ticket = 0
+  function takeATicketNumber(line){
+    ticket ++
+    line.push(ticket)
+    return (`Welcome. You are ticket number ${ticket}`)
+  }
+  return takeATicketNumber
+}
```
BrendaMichelle, line 234: Excellent! Only small thing - there's no need for the parentheses in the return on line 231. The same is true for other return statements in previous questions. This would work just fine:

    return `Welcome. You are ticket number ${ticketNumber}.`

BrendaMichelle (commented):
Nice work!

PR #2, opened 2020-12-10
JS Fundamentals 

wickedvic:
1-5 done. trouble with the others

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,69 +1,100 @@
 // ***** Question 1 *****
-// *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+let counter = 1
+counter = 2
+// *** Uncomment the lines below to test
+console.log("%cQuestion 1", "color: red") 
+
+console.log(counter) 
+// => 2
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
+const name = "Raffy"
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
-
-// console.log(name)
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1,69 +1,100 @@
...
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+console.log(counter) 
+// => 2
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
+const name = "Raffy"
```
BrendaMichelle, line 14: Great! The key here was using `const` because it doesn't allow us to reassign our variable

index.js
```diff
@@ -1,69 +1,100 @@
...
+
+function sameSameButDifferent(num, maybeNum){
+  if (num === maybeNum)
+    return "same same"
+  else if (num == maybeNum)
+    return "same same (but different)"
+  else
+    return "different"
+}
+
```
BrendaMichelle, line 56: This will work, but just a couple notes:

- for if/else statements, it's better to use curly parentheses to separate out each condition for readability/flexibility:

  `if (num === maybeNum) {
    return "same same"
  } else if (num == maybeNum) {
  } else {
    return "different"
  }`
- Also, you could also convert the string to a number with parseInt to avoid ==:

index.js
```diff
@@ -1,69 +1,100 @@
...
 
 
 
 // ***** Question 6 *****
+
+// function printNamesAndPhones(users){
+//   // i got stuck on this one
+//   // ik its using something with for each. but i kept getting a syntax error
+// }
+
```
BrendaMichelle, line 97: No worries. Next time, leave the code in and just comment it and I may be able to address the issue! The body should look like this:

 ```javascript
    console.log(user.name)
    console.log(`Cell: ${user.phones.cell}`)
    console.log(`Office: ${user.phones.office}`)
  })
```

BrendaMichelle:
Nice work. Left a few comments. Take a look and play around with the solution and see if you can figure out how the others work.

PR #3, opened 2020-12-10
sweet

escohen115:


index.js
```diff
@@ -1,88 +1,97 @@
...
-// ***** Question 3 *****
-// *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
+function drinkWater(currentThirstLevel){
+  console.log("Man I sure am thirsty")
+  currentThirstLevel --
+  console.log( "Ahh that hits the spot")
+  return(currentThirstLevel) 
+}
```
BrendaMichelle, line 21: Nice work! You don't need parenthesis for the return statement btw. This will work just fine, too:

`return currentThirstLevel`

index.js
```diff
@@ -1,88 +1,97 @@
...
+  if (num === maybeNum ){
+    return("same same")
+  }
+  else if (num == maybeNum){
+    return("same same (but different)")
+  }
+  else
+  {
+    return("different")
+  }
```
BrendaMichelle, line 38: - The key here was understanding the difference between `===` and `==`. You could also convert the string to a number with parseInt to avoid == :

- Only small thing - there's no need for the parentheses in the return on lines 30, 33, and 37.

index.js
```diff
@@ -1,88 +1,97 @@
...
-// // => "Duane"
+
+console.log("%cQuestion 6", "color: red")
+function printNameAndPhones (users){
+  for (let i = 0; i < users.length; i++)
+  {
+    console.log(users[i].name)
+    console.log(users[i].phones)
+  }
+}
```
BrendaMichelle, line 78: Almost! You should console log each phone number as well. It'll look like this:

```javascript
for (let i = 0; i < users.length; i++) {
    console.log(users[i].name)
    console.log(`Cell: ${users[i].phones.cell}`)
    console.log(`Office: ${users[i].phones.office}`)
}
```

index.js
```diff
@@ -99,24 +108,35 @@ function myMap(array, callback) {
...
 
 
 // ***** Callbacks - Question 2 *****
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 2", "color: red")
+console.log("%cCallbacks - Question 2", "color: red")
+
+function greet (name){
+  return (`Hi, ${name}`)
+}
```
BrendaMichelle, line 133: Excellent! Only small thing - there's no need for the parentheses in the return on line 132 here

BrendaMichelle:
Nice work! Left some comments

PR #4, opened 2020-12-10
Done, added some questions in comments

bstizzle:


index.js
```diff
@@ -1,88 +1,120 @@
...
+console.log(student1)
+// => { name: "Duane", grade: 92 }
+console.log("%c----------", "color: red")
 
 
 
 // ***** Question 6 *****
+//QUESTION FROM BEN
+//I tried to do this one writing a function() after the element => and put the logs inside that, but it didn't work
+//If I wanted the element to point to a function, do I have to define the function outside the forEach?
```
HyeokJungKim, line 90: When you use the `() => {}`, you're defining the function via an arrow function.
Whenever you use an arrow function, you don't have to use the `function` keyword.
But in regards to whether you want to use an anonymous function or a named function as a callback is a decision that you'll ultimately have to find

index.js
```diff
@@ -138,47 +175,65 @@ function nowServing(line) {
 const line = []
 
 // ***** Scope & Closures - Question 1 *****
+let place = 1;
 
+function takeATicketNumber(queue) {
+  queue.push(place);
+  return `Welcome. You are ticket number ${place++}`;
+  //that you can increment the variable INSIDE the string iterpolation is v cool - Ben
```
HyeokJungKim, line 183: Another cool thing is the difference between `place++` and `++place` !

index.js
```diff
@@ -138,47 +175,65 @@ function nowServing(line) {
...
 
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red") 
 
 // ***** Scope & Closures - Question 2 *****
+function ticketNumberGeneratorFunc() {
+  let count = 1; 
+  //this works, but I am a little confused as to why the count isn't
+  //resetting to 1 every time takeATicketNumberFunc is called
+  //is it because takeATicketNumberFunc is just the RETURN value of the GeneratorFunc (the nested function)?
```
HyeokJungKim, line 211: Exactly this!
This is closure at work.
Inside the scope of the `takeATicketNumberFunc, the `count` variable is accessible, which means you can read and both write that variable.

PR #6, opened 2020-12-10
Completed mini challenge

j-armah:


index.js
```diff
@@ -1,88 +1,136 @@
...
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red") 
+
+const name = "jeremy"
```

index.js
```diff
@@ -1,88 +1,136 @@
...
+  }
+  else {
+    if (num === parseInt(maybeNum)) {
+      return "same same (but different)"
+    }
+    else {
+      return "different"
+    }
+  }
+}
```
BrendaMichelle, line 70: Nice! This can be simplified to:

```javascript

function sameSameButDifferent (num, maybeNum) {
    if (num === maybeNum) {
      return "same same"
    }
    }
    else {
      return "different"
    }
}
```

You could also include an if statement around all three of the conditionals above if you want to ensure that num is a number using `typeof` or like you had above with `!Number.isNaN(num)`

index.js
```diff
@@ -1,88 +1,136 @@
...
-// updateGrade(student1, 92)
-// console.log(student1)
-// // => { name: "Duane", grade: 92 }
-// console.log("%c----------", "color: red")
+console.log("%cQuestion 5", "color: red")
+
+
+function updateGrade (studentObject, gradeNum) {
+  return studentObject["grade"] = gradeNum
+}
```
BrendaMichelle, line 94: nice! You could also use dot notation to access the studentObject's grade:

`studentObject.grade = gradeNum`

index.js
```diff
@@ -1,88 +1,136 @@
...
+  }
+]
+
+function printNameAndPhones (users) {
+  for (const user of users) {
+    console.log(user["name"])
+    // One with . op, one with bracket
+    console.log(`Cell: ${user.phones.cell}`)
+    console.log(`Office: ${user["phones"]["office"]}`)
+  }
```
BrendaMichelle, line 123: Nice! Getting practice with both dot notation and bracket notation, I see!

index.js
```diff
@@ -97,26 +145,34 @@ function myMap(array, callback) {
   return result
 }
 
+function triple (element) {
+  return element * 3
+}
+
+function greet (element) {
+  return `Hi, ${element}!`
+}
```
BrendaMichelle, line 154: well done

index.js
```diff
@@ -136,49 +192,73 @@ function nowServing(line) {
...
+function ticketNumberGeneratorFunc () {
+  let tickNum = 0
+  return function takeATicketNumberFunc (newLine) {
+    newLine.push(newLine.length + 1)
+    // tickNum = newLine.length doesn't work, why?
+    tickNum += 1
+
+    return `Welcome. You are ticket number ${tickNum}`
+  }
+}
```
BrendaMichelle, line 245: `tickNum = newLine.length` doesn't work because using the `newLine.length` way, the tickets don't get updated past a certain number. The test pushes two people into the line and gives them ticket 1 and 2. Then we serve the two front people (so the people who have tickets 1 and 2) and remove them from the line. Tickets 1 and 2 should be used up now, the next available ticket should be 3, and our line is empty. However, when we call `takeATicketNumberFunc(newLine)` for the last time, you use `newLine.length + 1` to compute the next ticket number, which is 1 because our line is empty.

BrendaMichelle (commented):
Nice work! Left some comments

PR #7, opened 2020-12-10
Done

k-will22:


index.js
```diff
@@ -1,88 +1,116 @@
...
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
-
-// console.log(name)
-// => "Raffy"
-// name = "Not Raffy" 
-// => TypeError
-// console.log("%c----------", "color: red") 
+console.log("%cQuestion 2", "color: red") 
+const = Keith
```
BrendaMichelle, line 17:
```suggestion
const = "Keith"
```

BrendaMichelle: Nice. The key here was using `const` because it doesn't allow us to reassign our variable. Just make sure your name is in quotes since it should be a string

index.js
```diff
@@ -1,88 +1,116 @@
...
-// // => "Man I sure am thirsty"
-// // => "Ahh that hits the spot"
-// // => 11
-// console.log("%c----------", "color: red") 
+function drink_water(current_thirst_level) {
+  console.log("Man I sure am thirsty")
+  current_thirst_level -= 1 
+  console.log("Ahh that hits the spot")
+  return current_thirst_level
+}
```
BrendaMichelle, line 35: Well done. For Javascript, it's standard to use camel case for function and variable names. So instead of `current_thirst_level`, name it `currentThirstLevel`

index.js
```diff
@@ -1,88 +1,116 @@
...
+
+function sameSameButDifferent(num, maybeNum) {
+  if num === maybeNum 
+    return "same same"
+  else if num == maybeNum 
+    return "same same (but different)"
+  else 
+  return "different"
+}
+console.log(sameSameButDifferent(5, 5))
```
BrendaMichelle, line 56: Close!

- It's standard to use parenthesis and curly braces for conditionals in Javascript. The lack of parenthesis around your conditional statements will cause an error. And curly braces around the conditional blocks increase your code's readability, makes your code more extensible, and reduces the risk of errors

```javascript
  if (num === maybeNum) { 
    return "same same"
}
  else if (num == maybeNum) { 
}
  else {
  return "different"
}
```

- The key here was understanding the difference between `===` and `==`, so good job with that. You could also convert the string to a number with parseInt to avoid == :

`else if (num === parseInt(maybeNum))`

index.js
```diff
@@ -1,88 +1,116 @@
...
-// console.log("%cQuestion 5", "color: red")
-// const student1 = { name: "Duane", grade: 88 }
-// updateGrade(student1, 92)
-// console.log(student1)
-// // => { name: "Duane", grade: 92 }
-// console.log("%c----------", "color: red")
+
+function updateGrade(student, grade) {
+  student['grade'] = grade 
+}
```
BrendaMichelle, line 76: Nice! You can also use dot notation like so:
`student.grade = grade`

index.js
```diff
@@ -1,88 +1,116 @@
...
-// console.log("%c----------", "color: red") 
+console.log("%cQuestion 6", "color: red")
+
+function printNameAndPhones(user) {
+  for (let i = 0; i < user.length; i++)
+    console.log([i]['name'])
+    console.log(`Cell: ${[i]['phones']['cell']}`)
+    console.log(`Office: ${[i]['phones']['office']}`)
+  }
+}
```
BrendaMichelle, line 96: close! If you want to use the for loop, it should look like this:

```javascript
for (let i = 0; i < user.length; i++) {
    console.log(user[i]['name'])
    console.log(`Cell: ${user[i]['phones']['cell']}`)
    console.log(`Office: ${user[i]['phones']['office']}`)
}
```

- You can also use dot notation instead of bracket notation:
```javascript
for (let i = 0; i < user.length; i++) {
    console.log(user[i].name)
    console.log(`Cell: ${user[i].phones.cell`)
    console.log(`Office: ${user[i].phones.office`)
}
```
____

For iterating over arrays, Javascript has a couple other methods that make it a little easier.
- Here's how it would look with for..of:
```javascript
  for (let user of users) {
    console.log(user.name);
  }
```
- Here's how it would look with forEach:
```javascript
    console.log(user.name);
  })
```

index.js
```diff
@@ -100,23 +128,31 @@ function myMap(array, callback) {
...
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 2", "color: red")
+console.log("%cCallbacks - Question 2", "color: red")
 
-// console.log(myMap(["Raffy", "Chase"], greet)) 
-// // => ["Hi, Raffy!", "Hi, Chase!"]
-// console.log("%c----------", "color: red") 
+function greet(name1, name2) {
+  console.log(`"Hi, ${name1}, "Hi, ${name2}!"`)
+}
```
BrendaMichelle, line 151: Close! This one is going to be like your `triple` function. It only takes one argument at a time.

```javascript
function greet(name) {
    return `Hi, ${name}`
}
```

index.js
```diff
@@ -100,23 +128,31 @@ function myMap(array, callback) {
 // ***** Callbacks - Question 1 *****
 
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 1", "color: red")
+console.log("%cCallbacks - Question 1", "color: red")
+ 
+function triple(value) {
+  console.log(value * 3)
+}
```
BrendaMichelle, line 135: On the right track! This function should return instead of console.logging because myMap function needs a return value to push to the new array it's making. Right now, the triple function will return undefined

```javascript
function triple(value) {
  console.log(value * 3)
}
```
_______________________________
See below for an example of using console.log vs return

```javascript
function addTwoNums(num1, num2) {
     return num1 + num2
}

let sum = addTwoNums(10, 20) // value of sum will be 30
```


VS

```javascript
function addTwoNums(num1, num2) {
     console.log(num1 + num2)
}

let sum = addTwoNums(10, 20) // value of sum will be undefined
```

index.js
```diff
@@ -135,6 +171,12 @@ function nowServing(line) {
   return `Currently serving ${line.shift()}.`
 }
 
+function takeATicketNumber(line) {
+    let ticket = line.length
+    console.log(`"Welcome. You are ticket number ${ticket}"`)
+    ticket++
+}
+
```
BrendaMichelle, line 179: I like the way you're thinking! However in this case, this will only work in some cases. By using `line.length` to determine the ticket number, you will be re-using ticket numbers

BrendaMichelle (commented):
You're on the right track! I left a few comments. Be sure to test your code early and frequently

PR #8, opened 2020-12-10
Done

jonathanx111:
Not 100 percent sure about the answers I wrote for the scoping questions

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,88 +1,119 @@
 // ***** Question 1 *****
+let counter = 1 
+counter = 2
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+console.log(counter) 
+// => 2
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
+const name = "jonathan"
+// name = "Not raffy"
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red") 
 
-// console.log(name)
+console.log(name)
 // => "Raffy"
 // name = "Not Raffy" 
 // => TypeError
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red") 
 
 
 
 // ***** Question 3 *****
+function drinkWater(current_thirst_level) {
+  console.log("Man I sure am thirsty")
+  current_thirst_level -= 1
+  console.log("Ahh that hits the spot")
+  return current_thirst_level
+}
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
 
-// console.log(drinkWater(12))
-// // => "Man I sure am thirsty"
-// // => "Ahh that hits the spot"
-// // => 11
-// console.log("%c----------", "color: red") 
+console.log(drinkWater(12))
+// => "Man I sure am thirsty"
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1,88 +1,119 @@
 // ***** Question 1 *****
+let counter = 1 
+counter = 2
```
BrendaMichelle, line 3: 💯

index.js
```diff
@@ -1,88 +1,119 @@
...
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+console.log(counter) 
+// => 2
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
+const name = "jonathan"
```
BrendaMichelle, line 13: Nice! The key here was using `const` because it doesn't allow us to reassign our variable

index.js
```diff
@@ -1,88 +1,119 @@
...
 
 
 
 // ***** Question 3 *****
+function drinkWater(current_thirst_level) {
+  console.log("Man I sure am thirsty")
+  current_thirst_level -= 1
+  console.log("Ahh that hits the spot")
+  return current_thirst_level
+}
```
BrendaMichelle, line 34: Well done! For Javascript, it's standard to use camel case for function and variable names. So instead of `current_thirst_level`, name it `currentThirstLevel`

index.js
```diff
@@ -1,88 +1,119 @@
...
 // ***** Question 4 *****
+function sameSameButDifferent(num, maybeNum) {
+  if (num === maybeNum & typeof num === 'number' & typeof maybeNum === 'number') {
+    return 'same same'
+  } else if ( num === parseInt(maybeNum)) {
+    return 'same same (but different)'
+  } else {
+    return 'different'
+  }
+}
```
BrendaMichelle, line 55: nice! Great idea ensuring that `num` is even a number in the first place 💯

index.js
```diff
@@ -1,88 +1,119 @@
...
 
 // ***** Question 6 *****
+function printOut(userHash) {
+  console.log(userHash['name'])
+  console.log(`Cell: ${userHash['phones']['cell']}`)
+  console.log(`Office: ${userHash['phones']['office']}`)
+}
+
+function printNameAndPhones(users) {
+  users.forEach(printOut)
```
BrendaMichelle, line 96: ooo, nice work!! Instead of accessing values in the userHash using bracket notation, you could also use dot notation. So: `userHash['phones']['cell']}` would be `userHash.phones.cell`, for example

index.js
```diff
@@ -138,47 +174,69 @@ function nowServing(line) {
...
 // ***** Scope & Closures - Question 1 *****
-
+let i = 0
+let a = 1
+function takeATicketNumber(line) {
+  line.push(a)
+  i++
+  a++
+  return `Welcome. You are ticketNumber${i}`
+}
```
BrendaMichelle, line 184: This works, but it can be simplified!

```suggestion
let i = 0
function takeATicketNumber(line) {
  i++
  line.push(i)
  return `Welcome. You are ticketNumber${i}`
}
```

index.js
```diff
@@ -138,47 +174,69 @@ function nowServing(line) {
...
+function ticketNumberGeneratorFunc() {
+  let i = 0
+  let a = 1
+  return function takeATicketNumber(line) {
+    line.push(a)
+    i++
+    a++
+    return `Welcome. You are ticketNumber${i}`
+  }
+}
```
BrendaMichelle, line 218: This is a tricky, advanced one! Nice work getting it! A higher order function is a function that accepts and/or returns another function. So this works, too, but can be simplified similarly to the previous question
```suggestion
function ticketNumberGeneratorFunc() {
  let i = 0
  return function takeATicketNumber(line) {
     i++
    line.push(i)
    return `Welcome. You are ticketNumber${i}`
  }
}
```

BrendaMichelle (commented):
Great work, this is really well done! Left some comments

PR #9, opened 2020-12-10
Johnson - js mini challenge

jyliu17:


index.js
```diff
@@ -1,42 +1,60 @@
...
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
 
-// console.log(name)
+const name = "Johnson"
```
BrendaMichelle, line 19: Nice! The key here was using `const` because it doesn't allow us to reassign our variable

index.js
```diff
@@ -1,42 +1,60 @@
...
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
 
-// console.log(drinkWater(12))
+function drinkWater(currentThirstLevel) {
+  console.log("Man I sure am thirsty")
+  console.log("Ahh that hits the spot")
+  console.log(currentThirstLevel - 1) 
+}
```
BrendaMichelle, line 38: Almost! 
- So your answer is not actually updating the value of `currentThirstLevel`. You need to add this to update it:
```javascript
 currentThirstLevel = currentThirstLevel - 1
```
- Another note, you are not returning anything from the function. You are just outputting the value to the console. `return` and `console.log` are not interchangeable. See this thread: [https://flatiron-school.slack.com/archives/C01DHCJ5QEM/p1607624963006700](https://flatiron-school.slack.com/archives/C01DHCJ5QEM/p1607624963006700)

index.js
```diff
@@ -1,42 +1,60 @@
...
-
+console.log("%cQuestion 4", "color: red") 
+
+function sameSameButDifferent(num, maybeNum) {
+  if (num == maybeNum)
+  return "same same"
+  if (num === parseFloat(maybeNum))
+  return "same same (but different)"
+  return "different"
+}
```
BrendaMichelle, line 57: Close! 
- The first conditional statement should include `===` instead of `==` so that javascript compares the values as well as the data type of each value
- It's standard in Javascript to use curly braces around conditional blocks

```suggestion
function sameSameButDifferent(num, maybeNum) {
  if (num === maybeNum) {
    return "same same"
  }
  else if (num === parseFloat(maybeNum)){
    return "same same (but different)"
  }
  return "different"
}
```

index.js
```diff
@@ -46,43 +64,54 @@
...
-// const student1 = { name: "Duane", grade: 88 }
-// updateGrade(student1, 92)
-// console.log(student1)
+
+
+console.log("%cQuestion 5", "color: red")
+
+function updateGrade(student, grade) {
+
+}
```
BrendaMichelle, line 81: The key here was understanding that objects are reference types. Object values are references, so when we mutate the student object inside updateGrade, the change will persist outside the function.

```javascript
function updateGrade(studentObject, gradeNumber) {
  studentObject.grade = gradeNumber
}
```

index.js
```diff
@@ -46,43 +64,54 @@
...
-//   { 
-//     name: "Liza", phones: { cell: "555-234-5678", office: "555-567-1234" }
-//   }
-// ]
-// printNameAndPhones(users)
+console.log("%cQuestion 6", "color: red")
+
+function printNameAndPhones(users) {
+  // users.forEach(printNameAndPhones)
+}
```
BrendaMichelle, line 98: I see you were trying to use `forEach` and pass in a callback function. I'm glad you knew the argument type that `forEach` takes. However, currently, this solution will run an infinite loop and probably error out.

```javascript
    console.log(user.name);
  })
```

index.js
```diff
@@ -100,23 +129,31 @@ function myMap(array, callback) {
 // ***** Callbacks - Question 1 *****
 
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 1", "color: red")
+console.log("%cCallbacks - Question 1", "color: red")
+
+function triple(threeTimes) {
+  console.log(threeTimes * 3)
+}
```
BrendaMichelle, line 136: Nice! Only issue is that you're console logging instead of returning. It's important for this function to return something because the myMap function is expecting back some return value. Here's the solution:

```javascript
function triple(num) {
  return threeTimes * 3
}
```

index.js
```diff
@@ -100,23 +129,31 @@ function myMap(array, callback) {
...
 
 
 // ***** Callbacks - Question 2 *****
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 2", "color: red")
+console.log("%cCallbacks - Question 2", "color: red")
+
+function greet(nameOne, nameTwo) {
+  console.log(`"Hi, ${nameOne}!", "Hi, ${nameTwo}!"`)
+}
```
BrendaMichelle, line 152: - same issue here as the previous question with needing a return value.
- myMap is expecting to be able to call this function passing in one argument at a time. So this should be: 
```javascript
function greet(nameOne) {
  console.log(`Hi, ${nameOne}!`)
}
```

index.js
```diff
@@ -141,15 +178,26 @@ const line = []
...
+function takeATicketNumber(line) {
+  
+  for(let i = 0; i < line.length; i++ ){
+    line.push(`${i+1}`);
+  }
+
+  return `Welcome. You are ticket number ${line}`
+  
+  
+}
```
BrendaMichelle, line 192: Hmm, whenever we call this function, we should only be updating the ticket number by one and pushing one ticket to the `line` array. Take another jab and then check out the solution and see if you can trace what the code is doing

BrendaMichelle (commented):
Keep up the hard work! Left some comments

PR #10, opened 2020-12-10
almost got em

amodttech:


index.js
```diff
@@ -1,88 +1,136 @@
...
+  if (num === maybeNum) {
+    console.log('same same')
+  } else if (num === parseInt(maybeNum, 10)){
+    console.log("same same (but different)")
+  } else {
+    console.log("different")
+  }
+
+  
+}
```
BrendaMichelle, line 64: 💯

index.js
```diff
@@ -1,88 +1,136 @@
...
+function sameSameButDifferent(num, maybeNum){
+  if (num === maybeNum) {
+    console.log('same same')
+  } else if (num === parseInt(maybeNum, 10)){
+    console.log("same same (but different)")
+  } else {
+    console.log("different")
+  }
+
+
```
BrendaMichelle, line 63:
```suggestion
```

index.js
```diff
@@ -1,88 +1,136 @@
...
-// printNameAndPhones(users)
-// // => "Duane"
+
+function printNameAndPhones(users){
+  users.forEach(function(element){
+    console.log(element.name)
+    console.log(`Cell: ${element.phones.cell}`)
+    console.log(`Office: ${element.phones.office}`)
+  })
+}
```
BrendaMichelle, line 113: well done!

index.js
```diff
@@ -139,26 +200,39 @@ const line = []
...
+  let line_counter = line.length
+  
+  // if (line.push(line_length) === 0){
+  //   line.push(1)
+  // }
+  
+
+  return `Welcome. You are number ${line.length} in line.`
+  line.unshift(line_counter)
+}
```
BrendaMichelle, line 214: Hmm, not exactly. You're never reaching line 213 because you `return` and finishing executing the function on line 212. However, that wouldn't solve the question, still. The answer does not necessarily relate to the length of the array because once we use a ticket number, we shouldn't be able to reuse that ticket again because the ticket number should increase regardless of how many people get in/get out of line. We are also interested in telling people what number ticket they have and not necessarily, how many people are in line before them. Reread the question and try again. Take a look at the solution and see if you can trace what it's doing

BrendaMichelle:
Well done! Almost there! Left some comments

PR #12, opened 2020-12-10
Done

ARElton:
In progress

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,9 +1,19 @@
+// Hi Michelle! I am still feeling quite shakey with objects and callbacks. 
+//I think I need time to play around more with labs, as yesterday I only got through the reading. 
+//I made it to question 6, but was pretty much stumped on and after that. 
+
+
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
+
+let counter = 1
+
+counter = 2 
+
 // console.log("%cQuestion 1", "color: red") 
 
 // console.log(counter) 
-// // => 2
+// //=> 2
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -100,14 +141,19 @@ function myMap(array, callback) {
...
-// // => [3,6,9,12,15]
+function triple() {
+    array * 3
+}
 
-// console.log(myMap([2,4,6,8], triple))   
-// // => [6,12,18,24]
-// console.log("%c----------", "color: red") 
+
+console.log(myMap([1,2,3,4,5], triple))
```
HyeokJungKim, line 151: If you're struggling with callbacks, I would take a look at this line and think about how the `triple` function is being passed in!

index.js
```diff
@@ -67,6 +101,13 @@
 // *** Uncomment the lines below to test
 // console.log("%cQuestion 6", "color: red")
 
+// function printNameAndPhones(users) {
+//   for (let i = 0; i < users.length; i++) {
```
HyeokJungKim, line 105: While a `for` loop is fine for iterating through an array, consider using a `forEach` to practice callbacks!
One possible solution for this would be:
```
users.forEach(function(user){
console.log(user.name)
console.log(user.phones.cell)
})
```

Remember that you can access keys in JavaScript objects via the dot notation, whether they're nested or not!

HyeokJungKim:
Don't worry if you weren't able to finish the challenge!
It's only been two days and the concept of callbacks is always tricky because functions behave differently than methods in Ruby.
You got this!

PR #13, opened 2020-12-10
Done! 

Ekenayy:
Wasn't able to complete the bonus question :(

### DIFF
--- a/index.js
+++ b/index.js
@@ -43,20 +43,30 @@ console.log("%c----------", "color: red")
 
 // ***** Question 4 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 4", "color: red") 
+console.log("%cQuestion 4", "color: red") 
 
-// console.log(sameSameButDifferent(5, 5)) 
-// // => "same same"
+function sameSameButDifferent(num, maybeNum) {
+  if (num === maybeNum) {
+  console.log("same same")
+  } else if (num == maybeNum) {
+  console.log("same same (but different)")
+  } else  {
+  console.log("different")
+  }
+}
 
-// console.log(sameSameButDifferent(123, "123")) 
-// // => "same same (but different)"
+console.log(sameSameButDifferent(5, 5)) 
+// => "same same"
 
-// console.log(sameSameButDifferent(5, 7)) 
-// // => "different"
+console.log(sameSameButDifferent(123, "123")) 
+// => "same same (but different)"
 
-// console.log(sameSameButDifferent(123, "122")) 
-// // => "different"
-// console.log("%c----------", "color: red") 
+console.log(sameSameButDifferent(5, 7)) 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,88 +1,123 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
-
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+console.log("%cQuestion 1", "color: red") 
+counter = 1
```
HyeokJungKim, line 4: I would get into the habit of using `let` or `const` whenever you declare a variable!
If you don't declare a variable using a keyword, it becomes a global variable in JS, which can potentially lead to bugs down the road!

index.js
```diff
@@ -1,88 +1,123 @@
...
+    name: "Duane", phones: { cell: "555-123-4567", office: "555-456-7890" }
+  },
+  { 
+    name: "Liza", phones: { cell: "555-234-5678", office: "555-567-1234" }
+  }
+]
+
+function printNameAndPhones(test_users) {
+  
+  for (i = 0; i < test_users.length; i++) {
```
HyeokJungKim, line 105: While using a `for` loop accomplishes array iteration, another solution you might want to consider is using `forEach`, since callbacks are going to be important to understand in Phase 3!
```
test_users.forEach(function(user){
   console.log(user.name)
   console.log(user.phones.cell)
})
```

index.js
```diff
@@ -135,50 +177,63 @@ function nowServing(line) {
   return `Currently serving ${line.shift()}.`
 }
 
-const line = []
+function takeATicketNumber(line) {
+  const line_number = line.push((line.length) + 1)
```
HyeokJungKim, line 181: Rather than pushing the `length` of the line into the `line`, I would push the `ticketCounter` variable in to the line.
So, if someone's the 100th customer, but the line is empty, we push their number into the array rather than 1

index.js
```diff
@@ -135,50 +177,63 @@ function nowServing(line) {
...
 // *** Uncomment the lines below to test
-// console.log("%cScope & Closures - Question 2", "color: red")
+console.log("%cScope & Closures - Question 2", "color: red")
+console.log("I am confuuuuused about this one")
+
+const newLine = []
+const takeATicketNumberFunc = function ticketNumberGeneratorFunc (line) {
+  var lineNumber = line.push((line.length) + 1)
+
+return `Welcome. You are ticket number ${lineNumber} `
```
HyeokJungKim, line 220: This function should return a function!
At the moment, it's returning a string.
Remember that in JS, we treat functions as objects, so we can return a function definition from a function.
So, the solution would look something like this:
```
function takeATicketNumberFunc(){
// ...
   return function(){
   }
})

PR #14, opened 2020-12-10
done

ConnorMul:


index.js
```diff
@@ -1,88 +1,119 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
-
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+console.log("%cQuestion 1", "color: red") 
+const counter = 2
```
HyeokJungKim, line 4: We want to be able to reassign the `counter` variable.
At the moment, because we are using `const`, we can't reassign the variable. :(

index.js
```diff
@@ -1,88 +1,119 @@
...
+  { 
+    name: "Duane", phones: { cell: "555-123-4567", office: "555-456-7890" }
+  },
+  { 
+    name: "Liza", phones: { cell: "555-234-5678", office: "555-567-1234" }
+  }
+]
+
+function  printNameAndPhones(arrOfHashes) {
+  for (i = 0; i < arrOfHashes.length; i++) {
```
HyeokJungKim, line 102: While using a `for` loop is fine, consider using `forEach` to practice callbacks.
Notice that you can also use the dot notation instead of the bracket notation!
```
arrOfHashes.forEach(function(user){
   console.log(user.name)
   console.log(user.phones.cell)
})
```

index.js
```diff
@@ -100,32 +131,39 @@ function myMap(array, callback) {
...
-function takeANumber(line, name) {
-  line.push(name)
+let num = 1
+
+function takeATicketNumber(line) {
 
-  return `Welcome, ${name}. You are number ${line.length} in line.`
+  line.push(num)
+  num++
+  return `Welcome. You are ticket number ${line.length}.`
```
HyeokJungKim, line 166: If you're customer number 100, and the line is empty, they'll receive that their ticket number is 1.
So, you should be returning the `num` in the interpolation rather than the `line.length`.
But doing that is going to affect the order of your code, so think about how you'd change that!

index.js
```diff
@@ -141,44 +179,51 @@ const line = []
...
-// const takeATicketNumberFunc = ticketNumberGeneratorFunc()
-// console.log(takeATicketNumberFunc(newLine))
-// // => `Welcome. You are ticket number 1`
+const newLine = []
+function ticketNumberGeneratorFunc(array) {
+  num = 1
+  array.push(num)
+  num++
+
+  return `Welcome. You are ticket number ${array.length}`
```
HyeokJungKim, line 211: You're still using a global variable with `newLine`.
In JS, functions can be the return value of a function, so the solution would look something like the following:
```
function ticketNumberGenerator(){
   // ...
   return function(){
   }
}
```

PR #15, opened 2020-12-10
Done

matthewogtong:
The bonus was difficult :(

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,88 +1,127 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+let counter = 1
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+counter = 2
+
+console.log("%cQuestion 1", "color: red") 
+
+console.log(counter) 
+// => 2
+console.log("%c----------", "color: red") 
+
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -141,33 +204,47 @@ const line = []
...
-// console.log("%cScope & Closures - Question 2", "color: red")
+console.log("%cScope & Closures - Question 2", "color: red")
 
-// const newLine = []
-// const takeATicketNumberFunc = ticketNumberGeneratorFunc()
-// console.log(takeATicketNumberFunc(newLine))
-// // => `Welcome. You are ticket number 1`
+function ticketNumberGeneratorFunc(takeATicketNumber){
+  ticket = 0
+  return takeATicketNumber(line)
```
HyeokJungKim, line 240: In JS, functions are considered to be objects, so function definitions can be returned from a function invocation!
So your solution might look something like this:
```
   // ...
   return function(){
   }
}
```

index.js
```diff
@@ -1,88 +1,127 @@
...
 
 // ***** Question 4 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 4", "color: red") 
+console.log("%cQuestion 4", "color: red") 
 
-// console.log(sameSameButDifferent(5, 5)) 
-// // => "same same"
+function sameSameButDifferent(num, maybeNum){
+  if (typeof(num) === 'number' && num === maybeNum) {
```
HyeokJungKim, line 54: Nice check to see if the `num` argument is a number!

PR #16, opened 2020-12-10
Done

antrubbo:
Gave it my best shot, stumped on the first closure question!

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,88 +1,130 @@
 // ***** Question 1 *****
-// *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+let counter = 1
+
+counter = 2
+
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,88 +1,130 @@
...
+console.log("%c----------", "color: red") 
 
 
 
 // ***** Question 3 *****
+
+function drinkWater(currentThirstLevel) {
+  console.log("Man I sure am thirsty");
+  console.log("Ahh that hits the spot");
+  return --currentThirstLevel;
```
HyeokJungKim, line 37: Nice use of the `--` operator!

index.js
```diff
@@ -139,26 +190,37 @@ const line = []
 
 // ***** Scope & Closures - Question 1 *****
 
+let newLine = []
+
+function takeATicketNumber(line) {
+  for (const element of line) {
```
HyeokJungKim, line 196: I like that you're using the for ... of loop, but think about how this would scale as you add/remove people from the line!

PR #17, opened 2020-12-10
done

kelseamc:


index.js
```diff
@@ -1,88 +1,121 @@
...
-// console.log("%cQuestion 4", "color: red") 
+console.log("%cQuestion 4", "color: red") 
 
-// console.log(sameSameButDifferent(5, 5)) 
-// // => "same same"
+function sameSameButDifferent(num, maybeNum){
+  if (num === maybeNum){
+     console.log("same, same")
+  }
+  else if (num == parseInt(maybeNum)) {
```
HyeokJungKim, line 52: `==` coerces the `maybeNum` into the same datatype as `num`, so it's unnecessary to use `parseInt`

index.js
```diff
@@ -1,88 +1,121 @@
...
+  { 
+    name: "Duane", phones: { cell: "555-123-4567", office: "555-456-7890" }
+  },
+  { 
+    name: "Liza", phones: { cell: "555-234-5678", office: "555-567-1234" }
+  }
+]
+
+function printNameAndPhones(array) {
+  for (let i = 0; i < array.length; i++){
```
HyeokJungKim, line 103: While using a `for` loop does accomplish array iteration, consider using a `forEach` to practice with callbacks!

index.js
```diff
@@ -136,35 +175,48 @@ function nowServing(line) {
...
+console.log("%c----------", "color: red") 
 
 // ***** Scope & Closures - Question 2 *****
 // *** Uncomment the lines below to test
-// console.log("%cScope & Closures - Question 2", "color: red")
+console.log("%cScope & Closures - Question 2", "color: red")
 
 // const newLine = []
+
+console.log("IDK what this question is even asking lol")
```
HyeokJungKim, line 219: lol
This function wants you to return a function that uses closure, so the solution would look something like this:
```
   // ...
   return function(){
   }
}
```

PR #19, opened 2020-12-10
Done

luistjr:


index.js
```diff
@@ -1,81 +1,125 @@
...
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
 
-// console.log(name)
+const name = "Raffy"
+
+console.log("%cQuestion 2", "color: red")
```

index.js
```diff
@@ -1,81 +1,125 @@
...
-// // => "Man I sure am thirsty"
-// // => "Ahh that hits the spot"
-// // => 11
-// console.log("%c----------", "color: red") 
+function drinkWater (thirst_level) {
+    console.log("Man I sure am thirsty")
+    console.log("Ahh that hits the spot")
+    console.log(thirst_level - 1)
+}
+
```
BrendaMichelle, line 38: Almost! Just a few notes:

- In Javascript, it's standard to use camel case for variable and functions names. So instead of `thirst_level`, the variable would be `thirstLevel`

- In order to actually change the value of `thirst_level`, you have to do this:

```javascript
thirstLevel = thirstLevel - 1
```
- Your function should explicitly return the updated thirst_level (there are no implicit returns in Javascript), so after line 36, you should have:
```
return thirstLevel
```

index.js
```diff
@@ -1,81 +1,125 @@
...
-// console.log(sameSameButDifferent(123, "122")) 
-// // => "different"
-// console.log("%c----------", "color: red") 
+  return "same same (but different)"
+
+} else { 
+
+  return "different"
+}
+}
```
BrendaMichelle, line 64: Good! Just work on your spacing and indentation - any time you have a new set of curly parentheses {} you should indent one level deeper 🙂

index.js
```diff
@@ -1,81 +1,125 @@
...
-// console.log(student1)
-// // => { name: "Duane", grade: 92 }
-// console.log("%c----------", "color: red")
+console.log("%cQuestion 5", "color: red")
+
+function updateGrade (student, grade) {
+
+  student["grade"] = grade
+
+}
```
BrendaMichelle, line 89: Good! For accessing the student's grade, you could also use dot notation:
```javascript
student.grade = grade
```

index.js
```diff
@@ -1,81 +1,125 @@
...
-//   }
-// ]
-// printNameAndPhones(users)
+console.log("%cQuestion 6", "color: red")
+
+function printNameAndPhones(users) {
+
+  
+
+}
```
BrendaMichelle, line 106: A few ways to solve this one. 
- One is with a good ol' for loop:
```javascript
for(let i = 0, i < users.length, i++) {
    console.log(users[i].name);
    console.log(`Cell: ${user[i].phones.cell}`);
    console.log(`Office: ${user[i].phones.office}`);
}
```

However, Javascript has a couple other methods that make it a little easier for iterating over arrays.

- Here's how it would look with for..of:
```javascript
  for (let user of users) {
    console.log(user.name);
  }
```

- Here's how it would look with forEach:
```javascript
    console.log(user.name);
  })
```

index.js
```diff
@@ -100,14 +144,16 @@ function myMap(array, callback) {
 // ***** Callbacks - Question 1 *****
 
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 1", "color: red")
+console.log("%cCallbacks - Question 1", "color: red")
 
-// console.log(myMap([1,2,3,4,5], triple)) 
-// // => [3,6,9,12,15]
+function triple
```
BrendaMichelle, line 149: This one would look like:
```javascript
function triple(num) {
     return num * 3
}
```

The myMap function takes two arguments. The array and a function. It iterates over each item in the array and calls the function on each of those items. Now that you have this solution, examine one line at a time what happens when we call `myMap([2,4,6,8], triple)`

BrendaMichelle:
Nice work! You're on the right track, Luis! Keep it up. Left some comments

PR #20, opened 2020-12-10
Barely done!

cswilik:


index.js
```diff
@@ -1,88 +1,127 @@
...
+  }
+  else {
+    return "different"
+  }
+    
+}
 
-// console.log(sameSameButDifferent(123, "123")) 
-// // => "same same (but different)"
+// == (doesn't check data type and value)
```
HyeokJungKim, line 61: This will try to coerce the data type!

index.js
```diff
@@ -1,88 +1,127 @@
...
+    name: "Duane", phones: { cell: "555-123-4567", office: "555-456-7890" }
+  },
+  { 
+    name: "Liza", phones: { cell: "555-234-5678", office: "555-567-1234" }
+  }
+]
+
+function printNameAndPhones(arrofUsers) {
+  let i = 0
+  while (i < arrofUsers.length) {
```
HyeokJungKim, line 110: While using a `while` loop can accomplish iterating through an array, consider using a `forEach` so that you can practice using a callback!

index.js
```diff
@@ -1,88 +1,127 @@
...
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
 
-// console.log(drinkWater(12))
-// // => "Man I sure am thirsty"
-// // => "Ahh that hits the spot"
-// // => 11
-// console.log("%c----------", "color: red") 
+function drinkWater(currentThirstLevel) {
```
HyeokJungKim, line 32: Be mindful what your function returns!
Just because it's logging a string does not mean that the function is returning something!

PR #21, opened 2020-12-10
done

thecodeplanner:


index.js
```diff
@@ -1,88 +1,122 @@
...
 // // => 11
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red") 
 
 
 
 // ***** Question 4 *****
+
+function sameSameButDifferent(num, maybeNum) {
+  convertNum = parseInt(maybeNum)
```
HyeokJungKim, line 52: When you run `sameSameButDifferent(123, "123"), this line will convert the "123" to 123, meaning that line 53 will evaluate to true.
Think about the differences between `==` and `===`!
Also, whenever you declare a variable, please make sure to use a variable declaration like `let` or `const`

index.js
```diff
@@ -1,88 +1,122 @@
...
-// console.log(sameSameButDifferent(123, "122")) 
+console.log(sameSameButDifferent(123, "122")) 
 // // => "different"
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red") 
 
 
 
 // ***** Question 5 *****
+function updateGrade (student, grade) {
```
HyeokJungKim, line 79: We want this function to alter the grade of the `student` object being passed in.
At the moment, we're nesting our information a little bit more and not altering the key in the `student` object.

index.js
```diff
@@ -1,88 +1,122 @@
...
+console.log(student1)
 // // => { name: "Duane", grade: 92 }
-// console.log("%c----------", "color: red")
+console.log("%c----------", "color: red")
 
 
 
 // ***** Question 6 *****
+
+function printNameAndPhones(users) {
```
HyeokJungKim, line 96: Consider using a `forEach` to iterate through the `users` array, rather than hard-coding the index values!

index.js
```diff
@@ -139,26 +184,31 @@ const line = []
 
 // ***** Scope & Closures - Question 1 *****
 
-
+let i = 1
+function takeATicketNumber(line) {
+  while (i < line.length) i++
```
HyeokJungKim, line 189: You have the right idea, but you only need to push `i` into the line once for each function execution!

PR #22, opened 2020-12-10
done

ashill99:


index.js
```diff
@@ -1,88 +1,118 @@
...
-
-// console.log(sameSameButDifferent(5, 5)) 
-// // => "same same"
+console.log("%cQuestion 4", "color: red") 
+
+function sameSameButDifferent(num, maybeNum) { 
+  const maybeNumInt = parseInt(maybeNum, 10)
+  if (Number.isInteger(num) && num === maybeNum)
+    return "same same"
+  if (num === maybeNumInt)
```
HyeokJungKim, line 54: The `==` operator coerces the `maybeNumInt` into the same dataType as num!
But, I like what you're doing here - being explicit with the conversion and then using the `===`

index.js
```diff
@@ -1,88 +1,118 @@
...
-//   }
-// ]
-// printNameAndPhones(users)
-// // => "Duane"
+console.log("%cQuestion 6", "color: red")
+
+
+function printNameAndPhones(users) 
+{
+  for (let i = 0; i < users.length; i++ )
```
HyeokJungKim, line 96: While using a `for` loop can accomplish iteration, consider using a `forEach` to get some practice in with callbacks!

index.js
```diff
@@ -100,24 +130,46 @@ function myMap(array, callback) {
...
+console.log("%c----------", "color: red") 
 
 
 // ***** Callbacks - Question 2 *****
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 2", "color: red")
+console.log("%cCallbacks - Question 2", "color: red")
+
+function greet(name) {
+  const namesArray = `Hi, ${name}!`
```
HyeokJungKim, line 165: While variable names technically don't matter, think about the data type for `namesArray`

index.js
```diff
@@ -141,44 +193,59 @@ const line = []
...
+
+const otherNewLine = []
 
-// console.log(takeATicketNumber(line))
-// // => `Welcome. You are ticket number 1`
+function takeATicketNumber(line) {
+otherNewLine.push(1)
+line.push(otherNewLine.length)
+return `Welcome. You are ticket number ${otherNewLine.length}`
+newLine += 1
```
HyeokJungKim, line 204: This line won't be executed because it's after the `return`!

index.js
```diff
@@ -141,44 +193,59 @@ const line = []
...
 
-// const newLine = []
-// const takeATicketNumberFunc = ticketNumberGeneratorFunc()
-// console.log(takeATicketNumberFunc(newLine))
-// // => `Welcome. You are ticket number 1`
+// Quite confused by this bonus question tbh 
+// pls explain a bit thank you!!!!!
+
+function ticketNumberGeneratorFunc() {
+  return takeATicketNumber
```
HyeokJungKim, line 232: You're on the right path in that this function returns a function.
However, the `takeATicketNumber` is still using the global variable of `otherNewLine`.
Take a look at the solution in the solution branch, but the general idea is that JS as a language treats functions as objects, so your solution might look something like this:
```
  // ...
  return function(){
  }
}
```

PR #23, opened 2020-12-10
done

anberjam:
was not able to complete the bonus problem.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,88 +1,126 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+let counter = 1
+ counter = 2
+ console.log("%cQuestion 1", "color: red") 
+
+ console.log(counter) 
+ // => 2
+ console.log("%c----------", "color: red") 
+
+ 
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,88 +1,126 @@
...
 
 
 
 // ***** Question 4 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 4", "color: red") 
+ console.log("%cQuestion 4", "color: red") 
+
+ function sameSameButDifferent (num, maybeNum) {
+  if (num === maybeNum && typeof(num)=="number"){
```
HyeokJungKim, line 52: Love that you're checking to see if the data type of `num` is a number!

index.js
```diff
@@ -1,88 +1,126 @@
...
-//   },
-//   { 
-//     name: "Liza", phones: { cell: "555-234-5678", office: "555-567-1234" }
-//   }
-// ]
-// printNameAndPhones(users)
+ console.log("%cQuestion 6", "color: red")
+
+function printNameAndPhones(users){
+  for (let i=0; i < users.length ;i++){
```
HyeokJungKim, line 101: While you can use a `for` loop to accomplish array iteration, consider using `forEach` to practice the idea of callbacks!

index.js
```diff
@@ -136,49 +191,61 @@ function nowServing(line) {
...
-// console.log("%cScope & Closures - Question 2", "color: red")
 
-// const newLine = []
-// const takeATicketNumberFunc = ticketNumberGeneratorFunc()
-// console.log(takeATicketNumberFunc(newLine))
+
+
+ console.log("%cScope & Closures - Question 2", "color: red")
+
+ //function TakeATicketNumberFunc () {}
```
HyeokJungKim, line 232: I would take a look at the solution in the `solution` branch!
It's an interesting solution!

PR #24, opened 2020-12-10
Almost done!

jcpleitez95:


index.js
```diff
@@ -1,92 +1,131 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+console.log("%cQuestion 1", "color: red") 
+
+var counter = 1
```
HyeokJungKim, line 6: While `var` does let you change the data-type, I would say for consistency's sake, to try and use `let` or `const` for variable declaration!

index.js
```diff
@@ -1,92 +1,131 @@
...
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
 
-// console.log(drinkWater(12))
-// // => "Man I sure am thirsty"
-// // => "Ahh that hits the spot"
-// // => 11
-// console.log("%c----------", "color: red") 
+function drinkWater(currentThirstLevel) {
```
HyeokJungKim, line 34: Think about what your function returns!
Just because something is logged to the console does not mean that it's being returned from that function!

index.js
```diff
@@ -1,92 +1,131 @@
...
-//   { 
-//     name: "Liza", phones: { cell: "555-234-5678", office: "555-567-1234" }
-//   }
-// ]
-// printNameAndPhones(users)
-// // => "Duane"
+console.log("%cQuestion 6", "color: red")
+
+function printNameAndPhones(users){
+  users.forEach(user => {
```
HyeokJungKim, line 101: Nice use of an arrow function!

PR #25, opened 2020-12-10
done

kbotero201:


index.js
```diff
@@ -25,6 +31,16 @@
 // *** Uncomment the lines below to test
 // console.log("%cQuestion 3", "color: red") 
 
+
+function drinkWater(current_thirst_level) {
```
HyeokJungKim, line 35: Think about what this function returns! 
Be mindful that what a function logs is different from what the function returns!

index.js
```diff
@@ -75,6 +107,15 @@
 //     name: "Liza", phones: { cell: "555-234-5678", office: "555-567-1234" }
 //   }
 // ]
+
+function print_name_and_phones(users){
+  users.forEach(){
```
HyeokJungKim, line 112: Remember that `forEach` expects a callback function!

PR #26, opened 2020-12-10
forked?

alfredwang0707:


index.js
```diff
@@ -1,96 +1,139 @@
...
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
-
+console.log("%cQuestion 1", "color: red") 
+let counter = 1
+console.log(counter + 1) 
 // console.log(counter) 
 // // => 2
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red")
```
BrendaMichelle, line 8: The output was correct, however the question asked for you to reassign the value of counter to 2.

Ex).
```javascript
let counter = 1
counter = 2
console.log(counter) // outputting the value to the console
```

index.js
```diff
@@ -1,96 +1,139 @@
...
+console.log("%cQuestion 2", "color: red") 
+const name = "Alfred"
+console.log(name)
+// name = "Robert"
+// Uncaught TypeError: Assignment to constant variable.
 // => "Raffy"
 // name = "Not Raffy" 
 // => TypeError
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red")
```
BrendaMichelle, line 23: Great! The key here was using `const` because it doesn't allow us to reassign our variable

index.js
```diff
@@ -1,96 +1,139 @@
...
+  console.log("Ahh that hits the spot")
+  return currentThirstLevel
+} 
+// def drink_water(current_thirst_level)
+//   puts "Man I sure am thirsty"
+//   current_thirst_level -= 1
+//   puts "Ahh that hits the spot"
+//   current_thirst_level
+// end
+console.log("%c----------", "color: red")
```
BrendaMichelle, line 47: Nice 💯

index.js
```diff
@@ -1,96 +1,139 @@
...
 
-// console.log(sameSameButDifferent(5, 7)) 
+console.log(sameSameButDifferent(5, 7)) 
 // // => "different"
 
-// console.log(sameSameButDifferent(123, "122")) 
+console.log(sameSameButDifferent(123, "122")) 
 // // => "different"
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red")
```
BrendaMichelle, line 76: Nice job! The key here was understanding the difference between `===` and `==`. You could also convert the string to a number with parseInt to avoid == :

index.js
```diff
@@ -1,96 +1,139 @@
...
+function updateGrade(student, grade){
+  student['grade'] = grade
+  return student 
+}
+const student1 = { name: "Duane", grade: 88 }
+updateGrade(student1, 92)
+console.log(student1)
 // // => { name: "Duane", grade: 92 }
-// console.log("%c----------", "color: red")
+console.log("%c----------", "color: red")
```
BrendaMichelle, line 91: Great job! 💯

index.js
```diff
@@ -100,23 +143,44 @@ function myMap(array, callback) {
...
+
+// The value of the element
+// The index of the element
+// The Array object being traversed
+
+console.log(myMap([1,2,3,4,5], triple)) 
 // // => [3,6,9,12,15]
 
-// console.log(myMap([2,4,6,8], triple))   
+console.log(myMap([2,4,6,8], triple))
```
BrendaMichelle, line 167: You're close, but you returned the function triple instead of returning the calculated value. 
Think of the callback function in this case as a function that is applied to each item inside the array.

To fix this you should remove your console log, and then move your calculation n*3 to be returned. 
Example).  
```javascript
function triple(n){
  return n * 3
}
```

index.js
```diff
@@ -100,23 +143,44 @@ function myMap(array, callback) {
...
-// console.log(myMap(["Raffy", "Chase"], greet)) 
+console.log(myMap(["Raffy", "Chase"], greet)) 
 // // => ["Hi, Raffy!", "Hi, Chase!"]
-// console.log("%c----------", "color: red") 
+
+function greet(array){
+  console.log(`Hi,${array}!`)
+  return array
+}
+console.log("%c----------", "color: red")
```
BrendaMichelle, line 183: You made a similar error here in the previous question. I reccomend checking out [this slack thread]((https://flatiron-school.slack.com/archives/C01DHCJ5QEM/p1607624963006700) for more info

The correct answer is:
```javascript
function greet(nameString) {
  return `Hi, ${nameString}!`
}
```
Remember, this function is applied to each item in the array. So the value "Raffy" at index 0 becomes, "Hi, Raffy" at index 0 of the new array created in the myMap function

index.js
```diff
@@ -136,36 +200,47 @@ function nowServing(line) {
 }
 
 const line = []
+let ticketNumber = 0 
+function takeATicketNumber(line){
+  //let ticketNumber = 0  will start 0 every time
+   ticketNumber ++
+   line.push(ticketNumber)
+   return `Welcome. You are ticket number ${ticketNumber}.`
```
BrendaMichelle, line 208: Logical semantic naming, good use of camel case for variable and function names. Correct logic 💯

index.js
```diff
@@ -1,96 +1,139 @@
...
+]
+// def print_name_and_phones(users)
+//   users.each do |user_hash|
+//     puts user_hash[:name]
+//     puts "Cell: #{user_hash[:phones][:cell]}"
+//     puts "Office: #{user_hash[:phones][:office]}"
+//   end
+// end
+function printNameAndPhones(users){
+  users.forEach(function(userInfo){
```
HyeokJungKim, line 115: Nice use of a callback!

index.js
```diff
@@ -136,36 +200,47 @@ function nowServing(line) {
...
 
 // ***** Scope & Closures - Question 2 *****
 // *** Uncomment the lines below to test
-// console.log("%cScope & Closures - Question 2", "color: red")
+console.log("%cScope & Closures - Question 2", "color: red")
 
-// const newLine = []
+const newLine = []
 // const takeATicketNumberFunc = ticketNumberGeneratorFunc()
+// ticketNumberGeneratorFunc(){
```
HyeokJungKim, line 241: Take a look at the solution, but remember that in JS, functions are objects!
Which means it can be returned from a function invocation!

PR #29, opened 2020-12-11
Done

serkansahin55:


index.js
```diff
@@ -1,43 +1,64 @@
...
-
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+console.log("%cQuestion 1", "color: red") 
+var counter = 1
+var counter = 2
+console.log(counter) 
+// => 2
+console.log("%c----------", "color: red")
```
BrendaMichelle, line 8: Just a few notes:
- Stick with using `let` or `const` over `var`. See this [forum](https://softwareengineering.stackexchange.com/questions/274342/is-there-any-reason-to-use-the-var-keyword-in-es6) for a discussion about it.
- Once you declare a variable (`let var`), you don't have to redeclare it in order to reassign it. In fact, you will get an error if you try to redeclare a variable in the same scope using `let`

The best solution here is:
```javascript
let counter = 1
counter = 2
```

index.js
```diff
@@ -1,43 +1,64 @@
...
 
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red")
+function drinkWater(currentThirstLevel){
+  console.log("Man I sure am thirsty")
+  console.log("Ahh that hits the spot")
+  console.log(currentThirstLevel -= 1)
+}
```
BrendaMichelle, line 34: Nice! Don't forget to `return` the `currentThirstLevel` as specified in the question

```javascript
function drinkWater(currentThirstLevel){
  console.log(currentThirstLevel -= 1)
  return currentThirstLevel
}
```

index.js
```diff
@@ -1,43 +1,64 @@
...
+//     result = "same same"
+
+//   } else if () {
+//     result = "same same (but different)"
+//   } else {
+//     result = "different"
+//   }
+//   return result;
+// }
+
```
BrendaMichelle, line 59: You're close. Think about `===` and `==` and `parseInt`.

```javascript
  let result = "different";
  if(num === maybeNum) {
    result = "same same";
  }
  else if(num === parseInt(maybeNum){
    result = "same same (but different)";
  }
  return result;
}
```

index.js
```diff
@@ -65,24 +89,31 @@
...
-// printNameAndPhones(users)
-// // => "Duane"
+console.log("%cQuestion 6", "color: red")
+
+function print_name_and_phones(users) {
+  users.forEach(function (){})
+  console.log(name)
+  console.log(cell)
+  console.log(office)
+}
```
BrendaMichelle, line 99: Inside the callback, think about what code you would want to write to console log the key value pairs of just one object. forEach is going to call this anonymous call back function on each item in the `users` array.

```javascript
users.forEach(function (oneUserObject){
  console.log(oneUserObject.name)
  console.log(`Cell: ${oneUserObject.phones.cell}`)
  console.log(`Office: ${oneUserObject.phones.office}`)
})
```

BrendaMichelle (commented):
Hey Serkan! Look over my feedback and see if it makes sense and test it out. Look at the solution for the callbacks and the first scope/closures question and try to trace what's happening in each solution. I recommend writing out the code with pen and paper and writing down what the values of the variables are at each line of code starting from the tests

PR #33, opened 2021-01-11
Done

jmarthaller:
First time making a pull request. Hope I'm doing it right!

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,88 +1,125 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+var counter = 1;
+var counter = 2;
+
+console.log(counter) 
+// => 2
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red") 
 
-// console.log(name)
+
+const name = "Jacob"
+console.log(name)
 // => "Raffy"
 // name = "Not Raffy" 
 // => TypeError
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red") 
 
 
 
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
 
-// console.log(drinkWater(12))
-// // => "Man I sure am thirsty"
-// // => "Ahh that hits the spot"
-// // => 11
-// console.log("%c----------", "color: red") 
+function drinkWater(current_thirst_level) {
+  console.log("Man I sure am thirsty");
+  current_thirst_level--;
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,88 +1,125 @@
...
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+var counter = 1;
+var counter = 2;
+
+console.log(counter) 
+// => 2
+console.log("%c----------", "color: red")
```
rlc900, line 10: This is really good logic, but i just wanna warn you about the "var" keyword! "Var" is function scoped, which means that you could access this variable from other functions whereas if you use the "let" keyword, it secures your variable as whatever scope you declare it in! Nice job :)

index.js
```diff
@@ -1,88 +1,125 @@
...
+
 
-// console.log(name)
+const name = "Jacob"
+console.log(name)
 // => "Raffy"
 // name = "Not Raffy" 
 // => TypeError
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red")
```
rlc900, line 25: Awesome!

index.js
```diff
@@ -1,88 +1,125 @@
...
-// console.log(drinkWater(12))
-// // => "Man I sure am thirsty"
-// // => "Ahh that hits the spot"
-// // => 11
-// console.log("%c----------", "color: red") 
+console.log(drinkWater(12))
+// => "Man I sure am thirsty"
+// => "Ahh that hits the spot"
+// => 11
+console.log("%c----------", "color: red")
```
rlc900, line 44: For the parameter in this function, JS convention is camelcase :) so current_thirst_level would be currentThirstLevel! Don't worry about not getting these syntax changes right away - ruby and JS are so so different! Otherwise though nice work!

index.js
```diff
@@ -1,88 +1,125 @@
...
-// // => "different"
+console.log(sameSameButDifferent(5, 7)) 
+// => "different"
 
-// console.log(sameSameButDifferent(123, "122")) 
-// // => "different"
-// console.log("%c----------", "color: red") 
+console.log(sameSameButDifferent(123, "122")) 
+// => "different"
+console.log("%c----------", "color: red")
```
rlc900, line 73: All of your logic is correct! On line 55 tho, you can compare the same params as on line 53 👍

index.js
```diff
@@ -1,88 +1,125 @@
...
+const users = [ 
+  { 
+    name: "Duane", phones: { cell: "555-123-4567", office: "555-456-7890" }
+  },
+  { 
+    name: "Liza", phones: { cell: "555-234-5678", office: "555-567-1234" }
+  }
+]
+printNameAndPhones(users)
+// => "Duane"
```
rlc900, line 116: All correct also!! Just wanted to point out "userHash" (they're the same in ruby and JS, but in JS they're called objects)

index.js
```diff
@@ -97,26 +134,37 @@ function myMap(array, callback) {
...
+function greet(name) {
+  console.log(`Hi, ${name}!`)
+}
 
-// console.log(myMap(["Raffy", "Chase"], greet)) 
-// // => ["Hi, Raffy!", "Hi, Chase!"]
-// console.log("%c----------", "color: red") 
+console.log(myMap(["Raffy", "Chase"], greet)) 
+// => ["Hi, Raffy!", "Hi, Chase!"]
+console.log("%c----------", "color: red")
```
rlc900, line 167: 👍

rlc900 (commented):
Awesome work overall! I'd say just to work on practicing some syntax corrections but all of your logic was there!

PR #35, opened 2021-01-11
Done with challenge

Jkang91:
Done!

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,18 +1,21 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+let counter = 1
+counter = 2
+console.log(counter) 
+// => 2
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red") 
 
+const name = "June"
 // console.log(name)
 // => "Raffy"
 // name = "Not Raffy" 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -23,66 +26,97 @@
...
+  return (num)
+}
 
-// console.log(drinkWater(12))
+console.log(drinkWater(12))
 // // => "Man I sure am thirsty"
 // // => "Ahh that hits the spot"
 // // => 11
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red")
```
rlc900, line 42: 👍

index.js
```diff
@@ -23,66 +26,97 @@
...
+    return ("same same")
+  }
+  else if(num == maybeNum){
+    return ("same same (but different)")
+  }
+  else{
+    return ("different")
+  }
+}
+console.log(sameSameButDifferent(5, 5))
```
rlc900, line 61: This is all correct! You don't need the () to return things though! You could just return the string :)

index.js
```diff
@@ -23,66 +26,97 @@
...
+
+function updateGrade(name, grade){
+  return(`name: ${name}, grade: ${grade}`)
+}
+const student1 = { name: "Duane", grade: 88 }
+updateGrade(student1, 92)
+console.log(student1)
 // // => { name: "Duane", grade: 92 }
-// console.log("%c----------", "color: red")
+console.log("%c----------", "color: red")
```
rlc900, line 87: This was super super close, but we want to pass in the student object as parameter instead of `name`. Because we only want this function to update the grade, we would target that as `student.grade = grade` (`student.grade` targeting the grades value and updating it to `grade` that youre passing into the function). Nice work though!!

index.js
```diff
@@ -23,66 +26,97 @@
...
+  }
+]
+users.forEach(printNameAndPhones)
+
+function printNameAndPhones(users){
+  console.log(`name: ${users.name}`)
+  console.log(`Cell: ${users.phones.cell}`)
+  console.log(`Office: ${users.phones.office}`)
+}
+
```
rlc900, line 111: WOW nice job using the forEach outside the loop! If it's more comfortable for you to do it this way, then def stick to it!

rlc900 (commented):
This was awesome work! I would just work on practicing returning values without the (), but your logic was spot on!!

PR #39, opened 2021-01-11
Done

jyliu17:


index.js
```diff
@@ -1,88 +1,129 @@
...
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
 
-// console.log(name)
+const name = "Johnson" 
+console.log("%cQuestion 2", "color: red") 
+
+console.log(name)
```
rlc900, line 24: Nice job! 👍

index.js
```diff
@@ -1,88 +1,129 @@
...
+console.log("%cQuestion 3", "color: red") 
 
-// console.log(drinkWater(12))
+function drinkWater(currentThirstLevel) {
+  console.log("Man I sure am thirsty")
+  console.log("Ahh that hits the spot")
+  console.log(currentThirstLevel -= 1) 
+}
+
+drinkWater(12)
```
rlc900, line 42: This is correct - but if you want to `return` the value of `currentThirstLevel`, you can subtract one from the current thirst like you had, and then `return currentThirstLevel`!

index.js
```diff
@@ -1,88 +1,129 @@
...
-// console.log(student1)
+
+
+console.log("%cQuestion 5", "color: red")
+
+function updateGrade(student, grade) {
+  student.grade = grade
+  
+
+}
```
rlc900, line 87: Nice work!

index.js
```diff
@@ -1,88 +1,129 @@
...
+const users = [ 
+  { 
+    name: "Duane", phones: { cell: "555-123-4567", office: "555-456-7890" }
+  },
+  { 
+    name: "Liza", phones: { cell: "555-234-5678", office: "555-567-1234" }
+  }
+]
+printNameAndPhones(users)
```
rlc900, line 119: Almost there!! On line 108, you would have to pass in the callBack func as a parameter to `printNameAndPhones` because the `callBack` you're calling on line 109 is not the same `callBack` as line 102. But if you just pass that in to the function when you declare and invoke it, it will work! Nice job 👍

rlc900 (commented):
Very nice job! I would just work on solidifying the logic for the `printNamesAndPhones` func, but other than that everything was spot on! 👍

PR #42, opened 2021-01-11
dont know last one. but i will ask

julioaht:
ready for some feedback

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,7 @@
 // ***** Question 1 *****
+let counter = 1
+counter = 2
+
 // *** Uncomment the lines below to test
 // console.log("%cQuestion 1", "color: red") 
 
@@ -13,6 +16,8 @@
 // *** otherwise, the error will stop the rest of your code from running
 // console.log("%cQuestion 2", "color: red") 
 
+const name = "Julio"
+
 // console.log(name)
 // => "Raffy"
 // name = "Not Raffy" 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,4 +1,7 @@
 // ***** Question 1 *****
+let counter = 1
+counter = 2
```
BrendaMichelle, line 3: Nice 💯  `let` allows us to reassign our variables

index.js
```diff
@@ -13,6 +16,8 @@
 // *** otherwise, the error will stop the rest of your code from running
 // console.log("%cQuestion 2", "color: red") 
 
+const name = "Julio"
```
BrendaMichelle, line 19: Great! The key here was using `const` because `const` doesn't allow us to reassign our variable

index.js
```diff
@@ -34,6 +49,20 @@
...
+    return "same same"
+  }
+  else if (num == maybeNum){
+    return "same same (but different)"
+  }
+  else {
+    return "different"
+  }
+
+}
```
BrendaMichelle, line 64: Good! Another approach would be to convert the string to a number with parseInt to avoid `==` :

index.js
```diff
@@ -98,6 +142,9 @@ function myMap(array, callback) {
 }
 
 // ***** Callbacks - Question 1 *****
+function triple(arrlenght){
+  return arrlenght * 3
+}
```
BrendaMichelle, line 147: Good! Only thing is your parameter name `arrlenght` indicates that maybe this function takes in the length of an array as an argument. However, it should take in any number value. `myMap` will call this function with each element in an array (not the array's length). It's just semantics, but maybe a better name for the parameter here would be `num` or `element`, etc.

index.js
```diff
@@ -139,6 +191,13 @@ const line = []
 
 // ***** Scope & Closures - Question 1 *****
 
+let ticketNum = 0
+
+function takeATicketNumber(line) {
+  ticketNum += 1
+  line.push(ticketNum)
+  return `Welcome. You are ticket number ${ticketNum}`
+}
```
BrendaMichelle, line 200: Good!

PR #43, opened 2021-01-11
Done minus bonus question

DanielYankiver:


index.js
```diff
@@ -1,89 +1,131 @@
 // ***** Question 1 *****
+
+let counter = 1 
+counter = 2
```
BrendaMichelle, line 4: Good!
`let` allows us to reassign our variables

index.js
```diff
@@ -1,89 +1,131 @@
...
-// // => 2
-// console.log("%c----------", "color: red") 
+console.log(counter) 
+// => 2
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
+
+const name = "Daniel"
```
BrendaMichelle, line 16: Great! The key here was using `const`to declare the variable because it doesn't allow us to reassign the variable

index.js
```diff
@@ -1,89 +1,131 @@
...
+  if (num === maybeNum) {
+    return "same same"
+  }
+  else if (num == maybeNum) {
+    return "same same (but different)"
+  }
+  else {
+    return "different"
+  }
+}
```
BrendaMichelle, line 63: Nice! Another solution would be to convert the string to a number with parseInt to avoid `==` :

index.js
```diff
@@ -139,46 +190,70 @@ const line = []
...
 
+let ticketNumber = 0 
+
+function takeATicketNumber(line) {
+  ticketNumber += 1 
+  line.push(ticketNumber)
+  return `Welcome. You are ticket number ${ticketNumber}`
+}
+// *** HAD TO GET A LITTLE HELP FROM SOLUTION TAB FOR THIS ONE 
+// *** Forgot --> let ticketNumber = 0
```
BrendaMichelle, line 201: Ah yes, the trick here is declaring `ticketNumber` outside the function because the function has access to the outer scope

index.js
```diff
@@ -139,46 +190,70 @@ const line = []
...
+
+function ticketNumberGeneratorFunc() {
+  let number = 0 
+
+  return 'Welcome you are ticket number ${number}'
+}
+
+// *** Couldn't get this one! A little confused on this topic still!! 
+
+
```
BrendaMichelle, line 237: This one's a tricky bonus one! Check this Scrimba out [https://scrimba.com/scrim/cLgEmZUZ](https://scrimba.com/scrim/cLgEmZUZ) if you haven't already if you want to try it again. The key here is to learn about higher order functions and closures. Think of the `ticketNumberGeneratorFunc` as a function factory. It should return a function that is in charge of increasing the ticket number and pushing the new ticket number into the array.

index.js
DanielYankiver: Awesome! Will try that out next time!

index.js
DanielYankiver: Sounds good! I will definitely check it out and revise. Appreciate all of your comments and feedback!

PR #44, opened 2021-01-11
Done

jaimedavid01:
Done

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,65 +1,94 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+let counter = 1
+counter++
+
+console.log(counter) 
+//=> 2
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red") 
 
-// console.log(name)
-// => "Raffy"
-// name = "Not Raffy" 
-// => TypeError
-// console.log("%c----------", "color: red") 
+const name = "Jaime"
+console.log(name)
+//=> "Raffy"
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,88 +1,128 @@
...
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red") 
 
-// console.log(name)
-// => "Raffy"
-// name = "Not Raffy" 
-// => TypeError
-// console.log("%c----------", "color: red") 
+const name = "Jaime"
```

index.js
```diff
@@ -1,88 +1,128 @@
...
+console.log("%cQuestion 3", "color: red") 
+
+console.log(drinkWater(12))
+function drinkWater(thirstLevel){
+// => "Man I sure am thirsty"
+console.log("Mann I sure am thirsty")
+thirstLevel -= 1
+// => "Ahh that hits the spot"
+console.log("Ahh that hits the spot")
+console.log(thirstLevel)
```
BrendaMichelle, line 39: Nice! Just remember, this function should be returning the `thirstLevel` and returning a value is different than just console logging a value
```suggestion
return thirstLevel
```

BrendaMichelle: - If we don't have the `return` and we try to use this function for getting a value, it will not work as expected as demonstrated in the following example:

```javascript
```

```javascript
```

index.js
```diff
@@ -1,88 +1,128 @@
...
+// => 11
```
BrendaMichelle, line 40: Also, side note - remember to indent one tab in after you open up a block of code `{`

index.js
```diff
@@ -1,88 +1,128 @@
...
+console.log("%cQuestion 4", "color: red") 
+
+function sameSameButDifferent(num, maybeNum){
+  if (num === maybeNum)
+  return "same same"
+  else if (num === parseInt(maybeNum)) {
+    return "same same (but different)" 
+  } else{
+    return "different"
+  }
```
BrendaMichelle, line 58: Good! Remember the curlies
```suggestion
function sameSameButDifferent(num, maybeNum){
  if (num === maybeNum) {
      return "same same"
  }
  else if (num === parseInt(maybeNum)) {
    return "same same (but different)" 
  } 
  else {
    return "different"
  }
```

index.js
BrendaMichelle, line 39: Also, side note - remember to indent one tab in after you open up a block of code {

PR #45, opened 2021-01-11
Done

zbayard:


index.js
```diff
@@ -1,7 +1,8 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
+// let counter = 1
 // console.log("%cQuestion 1", "color: red") 
-
+// let counter = 2
```
BrendaMichelle, line 5: Once you declare a variable (`let counter`), you can't redeclare it.

```javascript
let counter = 1 // declaration & assignment ✅

counter = 2 // re-assignment ✅

let counter = 3 //🚫  will produce an error because we are attempting to redeclare (`let counter`) a variable was already declared (on the first line here) 🚫
```

index.js
```diff
@@ -12,19 +13,24 @@
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
 // console.log("%cQuestion 2", "color: red") 
-
+// const name = "Zach"
```
BrendaMichelle, line 16: Great! The key here was using `const` because it doesn't allow us to reassign our variable

index.js
```diff
@@ -35,16 +41,30 @@
...
+//     return "same same"
+  
+//   else if (num == maybeNum)
+    
+//     return "same same (but different)"
+  
+//   else 
+//     return "different"
+  
+// }
```
BrendaMichelle, line 58: Good. Don't forget your curlies!

```suggestion
function sameSameButDifferent(num, maybeNum){
   if (num === maybeNum) {    
        return "same same"
   }
   else if (num == maybeNum) {
        return "same same (but different)"
   }
   else {
        return "different"
   }
 }
```

BrendaMichelle: Another solution for the `else if` would be to convert the string to a number with parseInt to avoid == :

index.js
```diff
@@ -166,6 +217,17 @@ const line = []
...
+
+// function ticketNumberGeneratorFunc(){
+
+//       MY MIND IS BLOWN HERE!
+      // MY MIND IS BLOWN HERE!
+      // MY MIND IS BLOWN HERE!
+      // MY MIND IS BLOWN HERE!
+      // MY MIND IS BLOWN HERE!
+
+// }
```
BrendaMichelle, line 229: his one's a tricky bonus one! Check this Scrimba out [https://scrimba.com/scrim/cLgEmZUZ ](https://scrimba.com/scrim/cLgEmZUZ ) and try again. Don't worry if you don't get it right now; It's just a bonus and it involves advanced, tricky concepts and this is only your first day with JavaScript!

PR #46, opened 2021-01-11
Done

maifam:


index.js
```diff
@@ -1,88 +1,121 @@
...
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red") 
 
-// console.log(name)
+const name = "Mai"
```
BrendaMichelle, line 20: Great! The key here was using `const` because it doesn't allow us to reassign our variable

index.js
```diff
@@ -1,88 +1,121 @@
...
+console.log("%cQuestion 4", "color: red") 
+
+function sameSameButDifferent(num, maybeNum) {
+  if (num === maybeNum)
+    return "same same"
+  else if (num == maybeNum)
+    return "same same (but different)"
+  else 
+    return "different"
+}
```
BrendaMichelle, line 57: Nice! One note is that it's good to get into the habit of using the curly brackets because it's the standard and that's how most javascript developers will write their javascript code. Using the curly brackets will also make your code less error prone.

 Another solution would be to convert the string to a number with parseInt to avoid == :

index.js
```diff
@@ -138,47 +179,71 @@ function nowServing(line) {
...
+
+// function ticketNumberGeneratorFunc(){
+//   function takeATicketNumber(line) {
+//     ticket += 1
+//     line.push(ticket)
+//     console.log`Welcome. You are ticket number ${ticket}`
+//   }
+//   return takeATicketNumber
+// }
+
```
BrendaMichelle, line 232: So close!! Don't worry if you're shaky or not completely understanding this one. This was just an advanced, tricky, bonus one and you did so well, especially for this only being the first day with Javascript! 

The `ticket` variable is currently global. Check out the `Lexical scope` and `Closure` section of this document [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures) and see if you realize what you need to change to make this work.

maifam:
Thank you so much for your feedback, Michelle! :)

PR #47, opened 2021-01-11
Done

clairedeboer:


index.js
```diff
@@ -1,88 +1,141 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+counter = 1; 
+counter = 2;
```
BrendaMichelle, line 4: Although this will work, it's best practice to declare your variables with a keyword.
```suggestion
let counter = 1; 
counter = 2; 
```

index.js
```diff
@@ -1,88 +1,141 @@
...
+// => 2
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+const name = "Claire";
```
BrendaMichelle, line 16: Great! The key here was using `const` because it doesn't allow us to reassign our variable

index.js
```diff
@@ -1,88 +1,141 @@
...
-// // => "Man I sure am thirsty"
-// // => "Ahh that hits the spot"
-// // => 11
-// console.log("%c----------", "color: red") 
+const drinkWater = (current_thirst_level) => {
+  console.log("Man I sure am thirsty"); 
+  current_thirst_level -= 1; 
+  console.log("Ahh that hits the spot"); 
+  return current_thirst_level; 
+}
```
BrendaMichelle, line 36: uuu, arrow functions, nice! Loving it 💯 Also, remember the convention in JavaScript is to use camelCase for variable and function naming. So instead of `current_thirst_level`, you'll have `currentThirstLevel`

index.js
```diff
@@ -1,88 +1,141 @@
...
+const sameSameButDifferent = (num, maybeNum) => {
+  if (num === maybeNum) {
+    return "same same"
+  }
+  if (num == maybeNum) {
+    return "same same (but different)"
+  } else {
+    return "different"
+  }
+}
```
BrendaMichelle, line 59: Good! Another solution would be to convert the string to a number with parseInt to avoid == :

index.js
```diff
@@ -1,88 +1,141 @@
...
-// console.log("%cQuestion 5", "color: red")
-// const student1 = { name: "Duane", grade: 88 }
-// updateGrade(student1, 92)
-// console.log(student1)
-// // => { name: "Duane", grade: 92 }
-// console.log("%c----------", "color: red")
+const updateGrade = (student, grade) => {
+  student.grade = grade
+  return student
+}
```
BrendaMichelle, line 83: Good! Even if you don't return `student`, you will still be able to see the updated value of the `student` object outside of the function. From this [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions) page:

> Arguments are passed to functions by value. If the function changes the value of an argument, this change is not reflected globally or in the calling function. However, object references are values, too, and they are special: if the function changes the referred object's properties, that change is visible outside the function

index.js
```diff
@@ -1,88 +1,141 @@
...
-// console.log("%cQuestion 6", "color: red")
 
-// const users = [ 
+const printNameAndPhones = (users) => {
+  users.forEach ((user) => {
+    console.log(user.name)
+    console.log(`Cell: ${user.phones.cell}`)
+    console.log(`Office: ${user.phones.office}`)
+  })
+}
```
BrendaMichelle, line 103: Great job using these arrow functions!!

PR #48, opened 2021-01-11
done

jpersons24:
All done, that was fun. Hoping that i did everything correctly, hopefully the console wasn't lying to me.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,88 +1,123 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+console.log("%cQuestion 1", "color: red")
 
-// console.log(counter) 
+let counter = 1
+counter = 2
+
+console.log(counter) 
 // // => 2
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red")
 
-// console.log(name)
+const name = "Jake Persons"
+
+console.log(name)
 // => "Raffy"
 // name = "Not Raffy" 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,88 +1,123 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+console.log("%cQuestion 1", "color: red")
 
-// console.log(counter) 
+let counter = 1
+counter = 2
```

index.js
```diff
@@ -1,88 +1,123 @@
...
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red")
+
+const name = "Jake Persons"
```

index.js
```diff
@@ -1,88 +1,123 @@
...
+
+function sameSameButDifferent(num, maybeNum) {
+  if (num === maybeNum) {
+    return "same same"
+  } else if (num == maybeNum) {
+    return "same same but different"
+  } else {
+    return "different"
+  };
+}
```
BrendaMichelle, line 60: Nice! Another solution would be to convert the string to a number with parseInt to avoid == :

index.js
```diff
@@ -1,88 +1,123 @@
...
-// printNameAndPhones(users)
+console.log("%cQuestion 6", "color: red")
+
+function printNameAndPhones(userArr) {
+  for (i = 0; i < userArr.length; ++i) {
+    console.log(userArr[i].name)
+    console.log(userArr[i].phones.cell)
+    console.log(userArr[i].phones.office)
+  }
+}
```
BrendaMichelle, line 103: Good work! Javascript has a couple other methods for iterating over arrays that make it a little easier.

```javascript
  for (let user of users) {
    console.log(user.name);
  }
```

```javascript
    console.log(user.name);
  })

BrendaMichelle (commented):
Great work! The console loggers did not lie. 😉  Left a few comments.

PR #49, opened 2021-01-11
not completed, but close

markghaida:
not completed, but close...I also left comments

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,81 +1,140 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+
+let counter = 1 
+counter = 2
+
+console.log(counter) 
+// => 2
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red") 
 
-// console.log(name)
+const name = "Mark"
+console.log(name)
 // => "Raffy"
 // name = "Not Raffy" 
 // => TypeError
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red") 
 
 
 
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
 
-// console.log(drinkWater(12))
-// // => "Man I sure am thirsty"
-// // => "Ahh that hits the spot"
-// // => 11
-// console.log("%c----------", "color: red") 
+function drinkWater(currentThirstLevel) {
+
+  console.log("Man I sure am thirsty"); 
+  currentThirstLevel-- ;
+  console.log("Ahh that hits the spot");
+  return currentThirstLevel;
+
+}
+
+
+console.log(drinkWater(12))
+// => "Man I sure am thirsty"
+// => "Ahh that hits the spot"
+// => 11
+console.log("%c----------", "color: red") 
 
 
 
 // ***** Question 4 *****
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,81 +1,140 @@
...
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+
+let counter = 1 
+counter = 2
+
+console.log(counter) 
+// => 2
+console.log("%c----------", "color: red")
```
Dwire, line 11: NIce job using let to define your variable here.

index.js
```diff
@@ -1,81 +1,140 @@
...
+console.log("%cQuestion 2", "color: red") 
 
-// console.log(name)
+const name = "Mark"
+console.log(name)
 // => "Raffy"
 // name = "Not Raffy" 
 // => TypeError
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red")
```
Dwire, line 25: Nice use of const

index.js
```diff
@@ -1,81 +1,140 @@
...
+
+  console.log("Man I sure am thirsty"); 
+  currentThirstLevel-- ;
+  console.log("Ahh that hits the spot");
+  return currentThirstLevel;
+
+}
+
+
+console.log(drinkWater(12))
```
Dwire, line 43: Nice work. Good use of -- and remembering that we need to explicitly return values in JS.

As an additional exercise, how would this look as an arrow function

index.js
```diff
@@ -1,81 +1,140 @@
...
+  else if (num == maybeNum) {
+    return "same same (but different)";
+  }
+
+  else {
+
+    return "different";
+  }
+}
+
```
Dwire, line 70: Good. 
- If statement syntax ✅
- return statements ✅
- use === vs == ✅

index.js
```diff
@@ -1,81 +1,140 @@
...
+console.log("%cQuestion 5", "color: red")
+
+const student1 = { name: "Duane", grade: 88 }
+
+updateGrade(student1, 92)
+
+console.log(student1)
+
+console.log("%c----------", "color: red")
```
Dwire, line 106: Nice Job
- Note: be careful writing variables that start with a capital. The convention would be to lowercase the UpdatedGrade variable
- In js dot notation is another common way to interact with objects. It would be possible to re-write line 93 like this `student.grade = updatedGrade`

index.js
```diff
@@ -1,81 +1,140 @@
...
+    console.log(`${users[i]["name"]}`)
+    console.log(`Cell: ${users[i]["phones"]["cell"]}`)
+    console.log(`Office: ${users[i]["phones"]["office"]}`)
+  }
+
+  } 
+
+
+
+console.log("%cQuestion 6", "color: red")
```
Dwire, line 127: Looks good!

Alternatively, you could use dot notation instead of bracket notation like I mentioned above or the .forEach method instead of a for loop. BUT nothing here is wrong. Nice work.

index.js
```diff
@@ -87,6 +146,16 @@
...
 
+
+function triple(element){
+  
+  let tripledArray = [];
+  tripledArray.push(element * 3);
+  return tripledArray;
+// i keep getting separate arrays for each element...I know that must be fixed 
+// but for the sake of time im moving on...look forward to feedback :)
+}
```
Dwire, line 157: Looks good. I do not think we need to create an array inside of the triple function. I believe this code will result in an array of arrays [[11],[23],[45],[9],[12]]. We just need triple to return a value that myMap will then add to the array.

index.js
```diff
@@ -100,23 +169,33 @@ function myMap(array, callback) {
...
-// console.log("%c----------", "color: red") 
+
+function greet(name){
+
+  let nameArray = []
+  nameArray.push(`Hi, ${name}`)
+  console.log(nameArray);
+
+}
+
```
Dwire, line 193: Nice work. same comment as above. I believe we want greet to just return `Hi, ${name}` instead of a single element array.

index.js
```diff
@@ -141,24 +220,27 @@ const line = []
 
 
 // *** Uncomment the lines below to test
-// console.log("%cScope & Closures - Question 1", "color: red")
 
-// console.log(takeATicketNumber(line))
-// // => `Welcome. You are ticket number 1`
+// didnt have enough time, so I am just submitting :/
```
Dwire, line 224: You did a lot good work!

PR #50, opened 2021-01-11
Still working on it X(

nanalau315:
I couldn't get the last two questions done...sorry!

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,9 +1,11 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
 // console.log("%cQuestion 1", "color: red") 
-
+// let counter = 1
 // console.log(counter) 
+// counter = 2
 // // => 2
+// console.log(counter) 
 // console.log("%c----------", "color: red") 
 
 
@@ -12,6 +14,11 @@
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
 // console.log("%cQuestion 2", "color: red") 
+// const name = "Nana"
+// console.log(name)
+
+// name = "Nana Lau"
+// console.log(name)
 
 // console.log(name)
 // => "Raffy"
@@ -24,7 +31,14 @@
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
 // console.log("%cQuestion 3", "color: red") 
+// let currentThirstLevel = 12
 
+// function drinkWater(currentThirstLevel) {
+//   console.log(`Man I sure am thirsty`)
+//   currentThirstLevel -= 1
+//   console.log(`Ahh that hits the spot`)
+//   return currentThirstLevel
+// }
 // console.log(drinkWater(12))
 // // => "Man I sure am thirsty"
 // // => "Ahh that hits the spot"
@@ -36,6 +50,17 @@
 // ***** Question 4 *****
 // *** Uncomment the lines below to test
 // console.log("%cQuestion 4", "color: red") 
+// function sameSameButDifferent(num, maybeNum) {
+//   if (num === maybeNum) {
+//     return `same same`
+//   }
+//   else if (num == maybeNum) {
+//     return`same same (but different)`
+//   }
+//   else
+//     return `different`
+// }
+
 
 // console.log(sameSameButDifferent(5, 5)) 
 // // => "same same"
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,9 +1,11 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
 // console.log("%cQuestion 1", "color: red") 
-
+// let counter = 1
 // console.log(counter) 
+// counter = 2
 // // => 2
+// console.log(counter)
```
Dwire, line 8: NIce job using let to define your variable here.

index.js
```diff
@@ -12,6 +14,11 @@
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
 // console.log("%cQuestion 2", "color: red") 
+// const name = "Nana"
+// console.log(name)
+
+// name = "Nana Lau"
+// console.log(name)
```
Dwire, line 21: Nice use of const

index.js
```diff
@@ -24,7 +31,14 @@
...
 // console.log("%cQuestion 3", "color: red") 
-
+// let currentThirstLevel = 12
+
+// function drinkWater(currentThirstLevel) {
+//   console.log(`Man I sure am thirsty`)
+//   currentThirstLevel -= 1
+//   console.log(`Ahh that hits the spot`)
+//   return currentThirstLevel
+// }
```
Dwire, line 41: Nice work remembering that we need to explicitly return values in JS.

As an additional exercise, how would this look as an arrow function?

index.js
```diff
@@ -36,6 +50,17 @@
...
+// function sameSameButDifferent(num, maybeNum) {
+//   if (num === maybeNum) {
+//     return `same same`
+//   }
+//   else if (num == maybeNum) {
+//     return`same same (but different)`
+//   }
+//   else
+//     return `different`
+// }
```
Dwire, line 62: Good. 
- If statement syntax ✅
- return statements ✅
- use === vs == ✅

Dwire: might just be missing one bracket in the else part of the if statement

index.js
```diff
@@ -55,7 +80,12 @@
 // ***** Question 5 *****
 // *** Uncomment the lines below to test
 // console.log("%cQuestion 5", "color: red")
+
 // const student1 = { name: "Duane", grade: 88 }
+// function updateGrade(student, grade) {
+//   student.grade = grade
+// }
```
Dwire, line 87: nice job using JS's dot notation
remember that the return value of this function as written will be undefined. This may not be a problem if we don't need the return value. Just pointing it out.

index.js
```diff
@@ -75,8 +105,15 @@
 //     name: "Liza", phones: { cell: "555-234-5678", office: "555-567-1234" }
 //   }
 // ]
+// function printNameAndPhones(user) {
+//   for (let i = 0; i < user.length; i++) {
+//     console.log(user[i].name)
+//     console.log(user[i].phones.cell)
+//     console.log(user[i].phones.office)
+//   }
+// }
```
Dwire, line 114: Nice job using dot notation for objects. Alternatively, you could have used a .forEach instead of a for loop.

index.js
```diff
@@ -101,6 +138,9 @@ function myMap(array, callback) {
 
 // *** Uncomment the lines below to test
 // console.log("%cCallbacks - Question 1", "color: red")
+// function triple(element) {
+//   return element * 3
+// }
```
Dwire, line 143: looks good 👌

index.js
```diff
@@ -114,6 +154,10 @@ function myMap(array, callback) {
 // *** Uncomment the lines below to test
 // console.log("%cCallbacks - Question 2", "color: red")
 
+// function greet(element){
+//   const greet = "Hi"
+//   return (`${greet}, ${element}!`)
+// }
```
Dwire, line 160: Nice work 😎

index.js
```diff
@@ -135,30 +179,33 @@ function nowServing(line) {
...
 // *** Uncomment the lines below to test
-// console.log("%cScope & Closures - Question 1", "color: red")
+console.log("%cScope & Closures - Question 1", "color: red")
 
-// console.log(takeATicketNumber(line))
-// // => `Welcome. You are ticket number 1`
+function takeATicketNumber(line) {
+    line.push(line[line.length - 1] + 1)
+    console.log(`Welcome. You are ticket number ${line.length - 1}`)
+}
```
Dwire, line 191: Nice solution. 

The section was hoping to practice a little with scope. The intended answer would create a counter variable in the global scope which the function would then use to add people/numbers to the line.

Dwire (commented):
Great Job. Looks good.

PR #53, opened 2021-01-11
mini challenge

cambisca:
Need more familiarity.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,10 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+let counter = 1
+counter = 2 
+
+// console.log(counter, "color: red") 
+
 
 // console.log(counter) 
 // // => 2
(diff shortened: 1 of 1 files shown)


Dwire (commented):
Nice job using "let" and "const" to define variables in the correct context.

index.js
```diff
@@ -11,10 +15,13 @@
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+const name = "Cam"
+
```
Dwire, line 19: Nice job using "let" and "const" in the correct context to define variables.

index.js
```diff
@@ -25,6 +32,12 @@
 // *** Uncomment the lines below to test
 // console.log("%cQuestion 3", "color: red") 
 
+function drinkWater(currentThirstLevel){
+  console.log("Man I sure am thirsty")
+  console.log("Ahh that hits the spot")
+  currentThirstLevel
+}
```
Dwire, line 39: Nice work. Remember to use the keyword "return" on line 38 if you want the function to return that value.

Dwire: Also create a new line under line 36 where we can change the thirst level.

  currentThirstLevel -= 1
  return currentThirstLevel

index.js
```diff
@@ -37,6 +50,18 @@
...
+  if (num === maybeNum){
+    return "same same"
+  }
+  elsif (num == maybeNum){
+    return "same same (but different)"
+  }
+  else {
+    return "different"
+  }
+}
```
Dwire, line 63: Good. 
- If statement syntax ✅
- return statements ✅
- use === vs == ✅

PR #55, opened 2021-02-01
JS Fundamentals Mini-Challenge Completed

dsasse07:


index.js
```diff
@@ -1,88 +1,135 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
-
-// console.log(counter) 
+console.log("%cQuestion 1", "color: red") 
+let counter = 1
+counter = 2
```
Dwire, line 5: some comment

index.js
```diff
@@ -1,88 +1,135 @@
...
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
-
-// console.log(name)
+console.log("%cQuestion 2", "color: red") 
+const name = "Danny"
+console.log(name)
 // => "Raffy"
 // name = "Not Raffy" 
 // => TypeError
```
Dwire, line 20: cool

index.js
```diff
@@ -1,88 +1,135 @@
...
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red")
```
Dwire, line 22: Nice work using the "let" and "const" keywords in the correct context when defining JS variables.

index.js
```diff
@@ -1,88 +1,135 @@
...
+//   puts "Ahh that hits the spot"
+//   current_thirst_level
+// end
+
+function drinkWater(current_thirst_level) {
+  console.log("Man I sure am thirsty")
+  current_thirst_level--
+  console.log("Ahh that hits the spot")
+  return current_thirst_level
+}
```
Dwire, line 41: Great work.  Remember that snake case is the standard or norm for defining ruby variables but camel case is the preferred way in JS. 
ruby = current_thirst_level 
JS = currentThirstLevel

Bonus: how would you write the above function as an arrow function?

index.js
```diff
@@ -1,88 +1,135 @@
...
+      console.log("same same")
+      break
+    case (num == maybeNum):
+      console.log("same same (but different)")
+      break
+    default:
+      console.log("different")
+      break
+  }
+
```
Dwire, line 67: Nice demonstration of control flow outside of the standard if-else statement. Good use of === vs ==.

index.js
```diff
@@ -1,88 +1,135 @@
...
+const student1 = { name: "Duane", grade: 88 }
+
+function updateGrade(student, newGrade){
+  student.grade = newGrade
+}
+
+updateGrade(student1, 92)
+console.log(student1)
+// => { name: "Duane", grade: 92 }
+console.log("%c----------", "color: red")
```
Dwire, line 97: Nice work upgrading the student grade using object dot notation. 

index.js
```diff
@@ -1,88 +1,135 @@
...
+    console.log(user.name)
+    for (const number in user.phones){
+      let properName = number.charAt(0).toUpperCase() + number.slice(1)
+      console.log(`${properName}: ${user.phones[number]}`)
+    }
+  });  
+  }
+
+
+printNameAndPhones(users)
```
Dwire, line 125: Nice Work!  Looks like you took this one a step further and made it slightly more flexible than the ruby equivalent. If we added a "home" number to the array yours would be able to handle it where the ruby example would not. Love to see it.

index.js
```diff
@@ -100,23 +147,27 @@ function myMap(array, callback) {
...
 
 // ***** Callbacks - Question 2 *****
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 2", "color: red")
+console.log("%cCallbacks - Question 2", "color: red")
 
-// console.log(myMap(["Raffy", "Chase"], greet)) 
+const greet = name => `Hi, ${name}!`
+
+console.log(myMap(["Raffy", "Chase"], greet))
```
Dwire, line 168: Nice work using arrow functions to make your code extremely concise. Great use of the one place JS provides implicit return values (in a single like no bracket arrow function). 

Only comment would be that you wrote your arrow functions using slightly different syntax. Always try to keep some consistency when writing your code. You should be able to drop the parens on your triple function.

const triple = num => num * 3
const greet = name => `Hi, ${name}!`

index.js
```diff
@@ -141,44 +192,60 @@ const line = []
...
+    newLine.push(newTicketNumber)
+    newTicketNumber++
+  }  
+}
 
-// const newLine = []
-// const takeATicketNumberFunc = ticketNumberGeneratorFunc()
-// console.log(takeATicketNumberFunc(newLine))
-// // => `Welcome. You are ticket number 1`
+console.log(takeATicketNumberFunc(newLine))
```
Dwire, line 236: Great job using the global variable approach and then being able to refactor that to use a closure. 👏

Dwire (commented):
Nice work. You showed a number of different skills outside the scope of this lab. Arrow functions, switch statements, .slice, completing the bonus. Love to see it.

index.js
dsasse07: I think it would be like this?

```js
const drinkWater2 = (currentThirstLevel) => {
  currentThirstLevel--
  return currentThirstLevel
}
```

index.js
dsasse07: `student1` is assigned as an object and cannot be reassigned to a different data type. Arrays and objects differ from primitive data type however, in that their contents can still be modified.

index.js
dsasse07: Thanks!

index.js
dsasse07: Is either way (with or without parentheses) generally considered "more proper" that that I can practice developing that method?

PR #57, opened 2021-02-01
complete

mcardona9015:


index.js
```diff
@@ -1,88 +1,116 @@
...
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red") 
 
-// console.log(name)
+console.log(name)
 // => "Raffy"
 // name = "Not Raffy" 
 // => TypeError
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red")
```
Dwire, line 23: Nice work using the "let" and "const" keywords in the correct context when defining JS variables.

index.js
```diff
@@ -1,88 +1,116 @@
...
 
 
 
 // ***** Question 3 *****
+function drinkWater(currentThirstLevel) {
+  console.log("Man I sure am thirsty");
+  currentThirstLevel -= 1;
+  console.log("Ahh that hits the spot");
+  return currentThirstLevel
+}
```
Dwire, line 33: Great work. The syntax looks good, you remembered the need for a return statement. I even like your attention to detail in re-naming your variable from the ruby standard snake case to the js standard camel case.

index.js
```diff
@@ -1,88 +1,116 @@
...
 // ***** Question 4 *****
+function sameSameButDifferent(num, MaybeNum) {
+  if (num === MaybeNum) {
+      return "same same"
+  } else if (num == MaybeNum) {
+      return "same same (but different)"
+  } else {
+      return "different"
+  }
+}
```
Dwire, line 54: Great use of a JS if-else statement and return statements to handle this control flow. Nice job differentiating between == vs ===.

index.js
```diff
@@ -1,88 +1,116 @@
...
-// updateGrade(student1, 92)
-// console.log(student1)
-// // => { name: "Duane", grade: 92 }
-// console.log("%c----------", "color: red")
+console.log("%cQuestion 5", "color: red")
+const student1 = { name: "Duane", grade: 88 }
+updateGrade(student1, 92)
+console.log(student1)
+// => { name: "Duane", grade: 92 }
+console.log("%c----------", "color: red")
```
Dwire, line 83: Nice work upgrading the student grade using object dot notation. 

index.js
```diff
@@ -1,88 +1,116 @@
...
 
 
 // ***** Question 6 *****
+function printNameAndPhones(users) {
+  users.forEach(userHash => {
+    console.log(userHash.name)
+    console.log(`Cell: ${userHash.phones.cell}`)
+    console.log(`Office: ${userHash.phones.office}`)
+  });
+}
```
Dwire, line 94: Looks solid. Nice job transforming things we were comfortable with in Ruby to their JS equivalent.

- .each  (vs) .forEeach
- "#{}" (vs) `${}`

index.js
```diff
@@ -98,25 +126,30 @@ function myMap(array, callback) {
...
-// console.log("%c----------", "color: red") 
+console.log(myMap([2,4,6,8], triple))   
+// => [6,12,18,24]
+console.log("%c----------", "color: red") 
 
 
 // ***** Callbacks - Question 2 *****
+function greet(name) {
+  return `Hi, ${name}!`
+}
```
Dwire, line 146: Great Job, callback functions return the correct values.

index.js
```diff
@@ -138,47 +171,71 @@ function nowServing(line) {
...
+console.log("%cScope & Closures - Question 2", "color: red")
 
-// const newLine = []
-// const takeATicketNumberFunc = ticketNumberGeneratorFunc()
-// console.log(takeATicketNumberFunc(newLine))
-// // => `Welcome. You are ticket number 1`
+const newLine = []
+const takeATicketNumberFunc = ticketNumberGeneratorFunc()
+console.log(takeATicketNumberFunc(newLine))
+// => `Welcome. You are ticket number 1`
```
Dwire, line 227: Great job solving this problem one way and then being able to refactor your code and find a different solution using closures. Seeing multiple paths to an outcome is not easy.

Dwire (commented):
Nice job, your JS code looks great!

PR #58, opened 2021-02-01
Still working but almost 5pm so submitting my pull request now

mahzmahzmahz:
Hi Michelle, I'm working through the minichallenge. THis material is a bit overwhelming and confusing but its starting to click little by little. Thanks for all your help!

Alex

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,13 +1,17 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+//console.log("%cQuestion 1", "color: red") 
+let counter = 1
 
-// console.log(counter) 
+counter = 2 
+console.log(counter) 
 // // => 2
 // console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
+
+const name = "Alex"
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
@@ -22,6 +26,16 @@
 
 
 // ***** Question 3 *****
+
+function drinkWater(currentThirstLevel){
+  console.log("Man I sure am thirsty")
+  console.log("Ahh that hits the spot")
+  currentThirstLevel -= 1
+  return currentThirstLevel
+}
+
+
+
 // *** Uncomment the lines below to test
 // console.log("%cQuestion 3", "color: red") 
 
@@ -37,10 +51,20 @@
 // *** Uncomment the lines below to test
 // console.log("%cQuestion 4", "color: red") 
 
-// console.log(sameSameButDifferent(5, 5)) 
+function sameSameButDifferent(num, maybeNum){
+  if (num === maybeNum) {
+    return "same same"
+  } else if (num == maybeNum) {
+    return "same same (but different)"
+  } else {
+    return different
+  }
+}
+
+ console.log(sameSameButDifferent(5, 5)) 
 // // => "same same"
 
-// console.log(sameSameButDifferent(123, "123")) 
+console.log(sameSameButDifferent(123, "123")) 
 // // => "same same (but different)"
 
 // console.log(sameSameButDifferent(5, 7)) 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,13 +1,17 @@
...
 // // => 2
 // console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
+
+const name = "Alex"
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
```
Dwire, line 17: Nice work using the "let" and "const" keywords in the correct context when defining JS variables.

index.js
```diff
@@ -22,6 +26,16 @@
...
 // ***** Question 3 *****
+
+function drinkWater(currentThirstLevel){
+  console.log("Man I sure am thirsty")
+  console.log("Ahh that hits the spot")
+  currentThirstLevel -= 1
+  return currentThirstLevel
+}
+
+
```
Dwire, line 37: Great work. The syntax looks good, you remembered the need for a return statement. I even like your attention to detail in re-naming your variable from the ruby standard snake case to the js standard camel case.

I would double-check your order of operations.  You probably want to satisfy your thirstLevel before announcing that it "hits the spot".  I would run `currentThirstLevel -= 1` before running your second console.log. 
NOTE: The output of your function would actually not be affected by this change at all. It is just always a good practice to think about our execution order when writing JS.

index.js
```diff
@@ -37,10 +51,20 @@
...
+  if (num === maybeNum) {
+    return "same same"
+  } else if (num == maybeNum) {
+    return "same same (but different)"
+  } else {
+    return different
+  }
+}
+
+ console.log(sameSameButDifferent(5, 5))
```
Dwire, line 64: Great use of a JS if-else statement and return statements to handle this control flow. Nice job differentiating between == vs ===.

index.js
```diff
@@ -53,36 +77,48 @@
...
-// const student1 = { name: "Duane", grade: 88 }
-// updateGrade(student1, 92)
-// console.log(student1)
+ console.log("%cQuestion 5", "color: red")
+ const student1 = { name: "Duane", grade: 88 }
+ updateGrade(student1, 92)
+ console.log(student1)
 // // => { name: "Duane", grade: 92 }
-// console.log("%c----------", "color: red")
+ console.log("%c----------", "color: red")
```
Dwire, line 90: Nice work upgrading the student grade using object dot notation. 

index.js
```diff
@@ -53,36 +77,48 @@
...
-// ***** Question 6 *****
+// ***** Question 6 *****{
+function printNameAndPhones(users){
+  users.forEach(function (user){
+    console.log(`${user.name}`)
+    console.log(`Cell: ${user.phones.cell}`)
+    console.log(`Office: ${user.phones.office}`)
+      }
+    )
+  }
```
Dwire, line 102: Looks solid. Nice job transforming things we were comfortable with in Ruby to their JS equivalent.

- .each  (vs) .forEeach
- "#{}" (vs) `${}`

index.js
```diff
@@ -99,24 +135,32 @@ function myMap(array, callback) {
...
 // // => [6,12,18,24]
-// console.log("%c----------", "color: red") 
+ console.log("%c----------", "color: red") 
 
 
 // ***** Callbacks - Question 2 *****
+function greet(names){
+  return ` Hi, ${names}!`
+}
+
```
Dwire, line 157: Both of your callback functions return the correct values.

index.js
```diff
@@ -138,47 +182,66 @@ function nowServing(line) {
...
+ console.log("%cScope & Closures - Question 2", "color: red")
 
-// const newLine = []
-// const takeATicketNumberFunc = ticketNumberGeneratorFunc()
-// console.log(takeATicketNumberFunc(newLine))
-// // => `Welcome. You are ticket number 1`
+ const newLine = []
+ const takeATicketNumberFunc = ticketNumberGeneratorFunc()
+ console.log(takeATicketNumberFunc(newLine))
+ // => `Welcome. You are ticket number 1`
```
Dwire, line 233: Great job solving this problem one way and then being able to refactor your code and find a different solution using closures. Seeing multiple paths to an outcome is not easy.

Dwire (commented):
Nice work! Solid day one progress.

PR #59, opened 2021-02-01
need to study more

victordcordero:
I feel able to do the first part but I need to read a-lot more about scope and callbacks/closures.

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,6 +5,9 @@
 // console.log(counter) 
 // // => 2
 // console.log("%c----------", "color: red") 
+let counter = 1;
+counter = 2;
+console.log(counter) 
 
 
 // ***** Question 2 *****
@@ -19,7 +22,8 @@
 // => TypeError
 // console.log("%c----------", "color: red") 
 
-
+const name = "Raffy"
+console.log(name)
 
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -5,6 +5,9 @@
 // console.log(counter) 
 // // => 2
 // console.log("%c----------", "color: red") 
+let counter = 1;
+counter = 2;
+console.log(counter)
```
BrendaMichelle, line 10: Nice 💯  `let` allows us to reassign our variables

index.js
```diff
@@ -31,22 +35,47 @@
...
+//     return "Ahh that hits the spot"
+//   }
+// }
+
+function drinkWater(currentThirstLevel) {
+  console.log("Man, I sure am thirsty")
+  console.log("Ahh that hits the spot")
+  return currentThirstLevel - 1
+}
+console.log(drinkWater(12))
```
BrendaMichelle, line 52: Good! 💯 No need for conditionals in this function because this function should always subtract 1 from the currentThirstLevel and should always console log those two statements

index.js
```diff
@@ -31,22 +35,47 @@
...
+function sameSameButDifferent(num, maybeNum) {
+  if (num === maybeNum){
+  return "same same"
+  }
+  else if (num == maybeNum){
+    return "same same (but different)"
+  }
+  else return "different"
+  }
+
```
BrendaMichelle, line 67: Nice! Another possible solution would be to convert the string to a number with parseInt to avoid == :

index.js
```diff
@@ -55,27 +84,45 @@
...
 // // => { name: "Duane", grade: 92 }
 // console.log("%c----------", "color: red")
 
+// Write a function `updateGrade` that takes two arguments: a `student` (object), and a `grade` (number). Calling this function should update the grade property on the `student` object to be the value of the `grade` argument.
 
-
+function updateGrade(student, grade){
+student.grade = grade
+console.log(student.name, grade)
+}
```
BrendaMichelle, line 98: Love to see you testing with these console logs!

index.js
```diff
@@ -55,27 +84,45 @@
...
+const users = [ 
+  { 
+    name: "Duane", phones: { cell: "555-123-4567", office: "555-456-7890" }
+  },
+  { 
+    name: "Liza", phones: { cell: "555-234-5678", office: "555-567-1234" }
+  }
+]
+printNameAndPhones(users)
```
BrendaMichelle, line 125: I know callbacks are tricky, but well done with this problem. You are on the right track with callbacks!

BrendaMichelle (commented):
Well done overall! Let us know if you have any questions or if you need working through any of the problems you didn't finish. If you re-attempt, let us know & we'll take a look over it.

PR #60, opened 2021-02-01
done

jtray86:


index.js
```diff
@@ -1,88 +1,125 @@
...
-
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+console.log("%cQuestion 1", "color: red") 
+let counter = 1
+counter = 2
+console.log(counter) 
+// => 2
+console.log("%c----------", "color: red")
```
BrendaMichelle, line 8: Nice 💯  `let` allows us to reassign our variables

index.js
```diff
@@ -1,88 +1,125 @@
...
+  return currentThirstLevel
+
+ }
 
-// console.log(drinkWater(12))
-// // => "Man I sure am thirsty"
-// // => "Ahh that hits the spot"
-// // => 11
-// console.log("%c----------", "color: red") 
+console.log(drinkWater(12))
```
BrendaMichelle, line 37: Looks good!

BrendaMichelle: One way to condense this and not lose any readability would be to consolidate lines 32 and 33 to:
```js
return currentThirstLevel - 1
```

index.js
```diff
@@ -1,88 +1,125 @@
...
+  if (num === maybeNum){
+    return "same same"
+  }
+  else if (num == maybeNum){
+    return "same same (but different)"
+  }
+  else {
+    return "different"
+  }
+}
```
BrendaMichelle, line 60: Nice! Another solution would be to convert the string to a number with parseInt to avoid == :

index.js
```diff
@@ -100,23 +137,33 @@ function myMap(array, callback) {
...
 // ***** Callbacks - Question 2 *****
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 2", "color: red")
 
-// console.log(myMap(["Raffy", "Chase"], greet)) 
-// // => ["Hi, Raffy!", "Hi, Chase!"]
-// console.log("%c----------", "color: red") 
+function greet(element){
+  return `Hi, ${element}!`
+}
```
BrendaMichelle, line 160: Well done!

index.js
```diff
@@ -138,47 +185,69 @@ function nowServing(line) {
...
-
+const ticketNum = []
+
+function takeATicketNumber(line){
+   
+  let num = ticketNum.length + 1
+  line.push(num)
+  ticketNum.push(num)
+  return `Welcome. You are ticket number ${num}`
+}
```
BrendaMichelle, line 196: Nice approach! This can be simplified. You're using an array to store the ticket numbers that have been used, but you're never actually using any of those values. Instead, you're using the length of that array. What if you just use a variable holding a Number data type to keep track of which ticket number  you are currently on instead?

index.js
```diff
@@ -138,47 +185,69 @@ function nowServing(line) {
...
+  const newticketNum = []
+  
+  return(function(line){
+    let newnum = newticketNum.length + 1
+    line.push(newnum)
+    newticketNum.push(newnum)
+    return `Welcome. You are ticket number ${newnum}`
+  })
+
+}
```
BrendaMichelle, line 230: Also, good! I have the same comment for this one as in the first scopes & closure question

BrendaMichelle (commented):
Well done! Left some comments

PR #61, opened 2021-02-01
done.

kody-samaroo:


index.js
```diff
@@ -1,88 +1,125 @@
...
-// console.log("%cQuestion 1", "color: red") 
+console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+counter = 1
+counter = 2
+
+console.log(counter)
```
BrendaMichelle, line 8: This works, but it's best to get into the habit of using a variable declaration keyword such as `let` for this example

index.js
```diff
@@ -1,88 +1,125 @@
...
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
+
+function drinkWater(thirst) {
+  console.log("Man I sure am thristy")
+  thirst -= 1
+  console.log("Ahh that hits the spot")
+  return thirst
+}
```
BrendaMichelle, line 39: Good! One possible way to condense this and not lose any readability would be to consolidate lines 35 and 37 to:
```js
return thirst - 1
```

index.js
```diff
@@ -1,88 +1,125 @@
...
+  if (num === maybeNum) {
+    console.log("same same")
+  }
+  else if (num == maybeNum) {
+    console.log("same same (but different)")
+  }
+  else {
+    console.log("different")
+  }
+}
```
BrendaMichelle, line 62: Nice! The key here was understanding the difference between `===` and `==`. You could also convert the string to a number with parseInt to avoid == :
Nice! Another solution would be to convert the string to a number with parseInt to avoid == :

index.js
```diff
@@ -1,88 +1,125 @@
...
+
+const student1 = { name: "Duane", grade: 88 }
+function updateGrade(student, grade) {
+  student.grade = grade
+}
+
+updateGrade(student1, 92)
+console.log(student1)
+// => { name: "Duane", grade: 92 }
+console.log("%c----------", "color: red")
```
BrendaMichelle, line 92: 💯

index.js
```diff
@@ -1,88 +1,125 @@
...
+const users = [ 
+  { 
+    name: "Duane", phones: { cell: "555-123-4567", office: "555-456-7890" }
+  },
+  { 
+    name: "Liza", phones: { cell: "555-234-5678", office: "555-567-1234" }
+  }
+]
+printNameAndPhones(users)
+// => "Duane"
```
BrendaMichelle, line 116: Well done! 💯

- Here's how the iteration would look using another approach with `forEach`:
```javascript
    console.log(user.name);
  })
```

index.js
```diff
@@ -100,23 +137,31 @@ function myMap(array, callback) {
...
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 2", "color: red")
+console.log("%cCallbacks - Question 2", "color: red")
 
-// console.log(myMap(["Raffy", "Chase"], greet)) 
-// // => ["Hi, Raffy!", "Hi, Chase!"]
-// console.log("%c----------", "color: red") 
+function greet(name) {
+  return `Hi, ${name}!`
+}
```
BrendaMichelle, line 160: 💯

index.js
```diff
@@ -141,44 +186,63 @@ const line = []
...
 // *** Uncomment the lines below to test
-// console.log("%cScope & Closures - Question 1", "color: red")
+console.log("%cScope & Closures - Question 1", "color: red")
+
+let ticket = []
+function takeATicketNumber(line) {
+  ticket.push(line.length)
+  console.log(`Welcome. You are ticket number ${ticket.length}`)
+  line.push(line.length + 1)
+}
```
BrendaMichelle, line 197: This approach works! This can be simplified by replacing the `ticket` array with another data type. You're currently only using the `ticket` array for its length and not because of any value it stores, so this beats the purpose of using an array. What other data type could you change `ticket` to simplify this?

index.js
```diff
@@ -141,44 +186,63 @@ const line = []
...
+  let newTicket = 1
+
+  return function (line) {
+      console.log(`Welcome. You are ticket number ${newTicket}`)
+      line.push(newTicket)
+      newTicket++
+  }
+
+}
+const takeATicketNumberFunc = ticketNumberGeneratorFunc()
```
BrendaMichelle, line 232: Great work!

BrendaMichelle (commented):
Well done! Left some comments

PR #62, opened 2021-02-01
done minus bonus

shmuly11:


index.js
```diff
@@ -1,88 +1,116 @@
...
-// console.log("%c----------", "color: red") 
+console.log(counter) 
+// => 2
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
+const name = "shmuly"
 // *** after testing, comment the line causing the error back in
```
Dwire, line 15: Nice work using the "let" and "const" keywords in the correct context when defining JS variables.

index.js
```diff
@@ -1,88 +1,116 @@
...
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+function drinkWater(currentThirstLevel){
+  console.log("Man I sure am thirsty")
+  currentThirstLevel --
+  console.log("Ahh that hits the spot")
+  return currentThirstLevel
+}
+console.log("%cQuestion 3", "color: red")
```
Dwire, line 35: Great work. The syntax looks good, you remembered the need for a return statement. I even like your attention to detail in re-naming your variable from the ruby standard snake case to the js standard camel case.

index.js
```diff
@@ -1,88 +1,116 @@
...
 // ***** Question 4 *****
+function sameSameButDifferent(num, maybeNum){
+  if(num === maybeNum){
+    return "same same"
+  } else if(num == maybeNum){
+    return "same same (but different)"
+  } else {
+    return "different"
+  }
+}
```
Dwire, line 54: Great use of a JS if-else statement and return statements to handle this control flow. Nice job differentiating between == vs ===.

index.js
```diff
@@ -1,88 +1,116 @@
...
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 5", "color: red")
-// const student1 = { name: "Duane", grade: 88 }
-// updateGrade(student1, 92)
-// console.log(student1)
+console.log("%cQuestion 5", "color: red")
+const student1 = { name: "Duane", grade: 88 }
+updateGrade(student1, 92)
+console.log(student1)
```
Dwire, line 81: Nice work upgrading the student grade using object dot notation.

index.js
```diff
@@ -1,88 +1,116 @@
...
 
 
 // ***** Question 6 *****
+function printNameAndPhones(users){
+  for (let user in users){
+    console.log(users[user].name)
+    console.log(`Cell: ${users[user].phones.cell}`)
+    console.log(`Office: ${users[user].phones.office}`)
+  }
+}
```
Dwire, line 94: Looks solid. Nice job transforming things we were comfortable with in Ruby to their JS equivalent.

- .each  (vs) for in loop
- "#{}" (vs) `${}`

index.js
```diff
@@ -98,35 +126,41 @@ function myMap(array, callback) {
...
-// console.log("%c----------", "color: red") 
+console.log(myMap([2,4,6,8], triple))   
+// => [6,12,18,24]
+console.log("%c----------", "color: red") 
 
 
 // ***** Callbacks - Question 2 *****
+function greet (element){
+  return `Hi, ${element}!`
+}
```
Dwire, line 146: Nice work, your callback functions both look to be returning the correct values.

index.js
```diff
@@ -158,27 +192,39 @@ const line = []
...
-// // => `Welcome. You are ticket number 1`
+  return takeATicketNumber
 
-// console.log(takeATicketNumberFunc(newLine))
-// // => `Welcome. You are ticket number 2`
+}
+const newLine = []
+const takeATicketNumberFunc = ticketNumberGeneratorFunc()
+console.log(takeATicketNumberFunc(newLine))
+// => `Welcome. You are ticket number 1`
```
Dwire, line 216: Great job using a closure to solve this problem. Your pull request says that you did not do the Bonus. But this closure looks good to me. If you did not complete this on your own, don't sweat it, closures are tough and take practice to get comfortable with.

Dwire (commented):
Nice Work. Your code looks very good and it appears you are successfully transitioning Ruby concepts into new JS syntax. Great day 1!

PR #63, opened 2021-02-01
Done

BoltVanderhuge:


index.js
```diff
@@ -1,88 +1,121 @@
...
 
 
 
 // ***** Question 3 *****
+  function drinkWater(currentThirstLevel) {
+    console.log("Man I sure am thirsty")
+    currentThirstLevel -= 1 
+    console.log("Ahh that hits the spot")
+    return currentThirstLevel 
+  }
```
hasibulc, line 33: Great use of explicit return

index.js
```diff
@@ -1,88 +1,121 @@
...
+      return "same same"
+    }
+    else if (num == maybeNum) {
+      return "same same (but different)"
+    }
+    else {
+      return "different"
+    }
+
+  }
```
hasibulc, line 58: good job with === vs ==

index.js
```diff
@@ -98,25 +131,30 @@ function myMap(array, callback) {
...
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 1", "color: red")
+console.log("%cCallbacks - Question 1", "color: red")
 
-// console.log(myMap([1,2,3,4,5], triple)) 
+console.log(myMap([1,2,3,4,5], triple)) 
 // // => [3,6,9,12,15]
 
-// console.log(myMap([2,4,6,8], triple))   
+console.log(myMap([2,4,6,8], triple))
```
hasibulc, line 143: not sure if myMap was defined and then deleted right before the pull request, currently there is no callback function "myMap"

index.js
```diff
@@ -138,47 +176,60 @@ function nowServing(line) {
 const line = []
 
 // ***** Scope & Closures - Question 1 *****
-
+let tickynumber = 0
+  function takeATicketNumber(line){
+    tickynumber += 1
+    line.push(tickynumber)
+    return `Welcome. You are ticket number ${tickynumber}`
+  }
```
hasibulc, line 184: Good job using a second variable to keep track of current ticket number!

index.js
```diff
@@ -138,47 +176,60 @@ function nowServing(line) {
...
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red") 
 
 // ***** Scope & Closures - Question 2 *****
+  function ticketNumberGeneratorFunc() {
+    let tickynumber = 0
+    return function (line){
+      tickynumber += 1
+      line.push(tickynumber)
+      return `Welcome. You are ticket number ${tickynumber}`
```
hasibulc, line 212: good use of an anonymous function!

hasibulc:
Overall great job with the mini challenge. I would suggest using better variable names moving forward, ie: 'tickynumber'.

PR #66, opened 2021-02-01
submit

samhall330:


index.js
```diff
@@ -1,184 +1,249 @@
...
-// console.log("%c----------", "color: red") 
-
-
-
+// console.log("%cQuestion 2", "color: red")
+ 
+const name = "Sam"
+ 
+ 
+
```
Dwire, line 21: Nice work using the "let" and "const" keywords in the correct context when defining JS variables.

index.js
```diff
@@ -1,184 +1,249 @@
...
+ current_thirst_level -=1
+ console.log("Ahh that hits the spot.")
+ return current_thirst_level
+}
+ 
+console.log(drinkWater(12))
+ 
+ 
+ 
+
```
Dwire, line 37: Great work. Remember that snake case is the standard or norm for defining ruby variables but camel case is the preferred way in JS. 

ruby = current_thirst_level 
JS = currentThirstLevel

index.js
```diff
@@ -1,184 +1,249 @@
...
+   return "same same (but different)"
+   }
+ else {
+   return "different"
+   }
+ }
+ 
+ 
+ 
+console.log(sameSameButDifferent(5, 5))
```
Dwire, line 56: Great use of a JS if-else statement and return statements to handle this control flow. Nice job differentiating between == vs ===.

index.js
```diff
@@ -1,184 +1,249 @@
...
+ 
+// console.log("%c----------", "color: red")
+ 
+function updateGrade(student, updatedGrade){
+ student.grade = updatedGrade
+}
+ 
+const student1 = { name: "Duane", grade: 88 }
+updateGrade(student1, 92)
+updateGrade(student1, 67)
```
Dwire, line 84: Nice work upgrading the student grade using object dot notation. 

index.js
```diff
@@ -1,184 +1,249 @@
...
+function printNameAndPhones(users){
+for (let i = 0 < users.length; i++) {
+  console.log(user.name)
+  console.log(user.phones)
+}
+}
+// Uncaught SyntaxError: Unexpected token ')' -- can't figure this out! UGH.
+ 
+ 
+console.log(printNameAndPhones(users))
```
Dwire, line 120: Nice attempt. if we are going to use a for loop here then we do not have access to the variable `user` that we are trying to use in your for loop. Instead, we would have to use the full array `users` and the current index `i` to find the individual user.

for (let i = 0 < users.length; i++) {
  console.log(users[i].name)
  console.log(users[i].phones.cell)
  console.log(users[i].phones.office)
}

the other way we could do this would be by using a `forEach` method in which case we would define a `user` variable to use.

function printNameAndPhones(users){
  users.forEach (function(user) {
    console.log(user.name)
    console.log(user.phones.cell)
  })
}

index.js
```diff
@@ -1,184 +1,249 @@
...
-// console.log("%c----------", "color: red") 
-
-
+// console.log("%c----------", "color: red")
+ 
+function triple(number){
+ number * 3
+}
+ 
+
```
Dwire, line 177: Your code is super close. Just remember that in JS we need to be explicit with our returns. Currently, your `triple` function is returning `undefined` because we have not told it what to return.

try adding a return to line 174
function triple(number){
 return number * 3
}

Dwire (commented):
Nice work. a couple of things to work on with iteration and remembering to make sure our functions always `return` the correct value when needed. But overall great work and a solid day one of JS.

PR #67, opened 2021-02-01
best i could 

menachemmargolis:


index.js
```diff
@@ -1,81 +1,120 @@
...
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+ console.log("%cQuestion 3", "color: red") 
+
+ function drinkWater(current_thirst_level){
+  console.log ("Man I sure am thirsty")
+  current_thirst_level -= 1
+  console.log ("Ahh that hits the spot")
+  return current_thirst_level
+ }
```
Dwire, line 34: Great work transforming Ruby into JS. 

remember that the standard/norm for declaring variables in ruby is snake case and the standard in JS is camel case. 
To clean up this JS I would change the variable from `current_thirst_level ` to `currentThirstLevel`. 

NOTE: both work and do not affect the outcome of your function.

index.js
```diff
@@ -1,81 +1,120 @@
...
+  if (num === maybeNum){
+   return'same same'
+  }
+ else if  (num == maybeNum) {
+  return 'same same but different'
+  }
+  else{
+    return 'not same'
+  }
+}
```
Dwire, line 58: Great use of a JS if-else statement and return statements to handle this control flow. Nice job differentiating between == vs ===.

index.js
```diff
@@ -1,81 +1,120 @@
...
+
+ 
+ const student1 = { name: "Duane", grade: 88 }
+
+ function updateGrade(student, grade){
+  student.grade = grade 
+ }
+ updateGrade(student1, 92)
+
+ console.log(student1)
```
Dwire, line 85: Nice work upgrading the student grade using object dot notation.

index.js
```diff
@@ -1,81 +1,120 @@
...
+   console.log (`office ${value.phones.office}`)
+  })
+ 
+  
+}
+
+
+printNameAndPhones(users)
+
+
```
Dwire, line 117: Looks solid. Nice job transforming things we were comfortable with in Ruby to their JS equivalent.

- .each  (vs) .forEeach
- "#{}" (vs) `${}`

index.js
```diff
@@ -100,23 +139,31 @@ function myMap(array, callback) {
...
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 2", "color: red")
+ console.log("%cCallbacks - Question 2", "color: red")
 
-// console.log(myMap(["Raffy", "Chase"], greet)) 
+ function greet(dog){
+   return `hello i missed u ${dog}`
+ }
+
+ console.log(myMap(["Raffy", "Chase"], greet))
```
Dwire, line 164: Nice Job! Both of your callback functions return the correct values to the higher-order function that calls them.

index.js
```diff
@@ -137,6 +184,16 @@ function nowServing(line) {
...
+
+let numberInLine = 0
+
+
+
+function takeATicketNumber(line){
+  numberInLine += 1
+  line.push(numberInLine)
+  return `hello ticket number${numberInLine} you are next`
+}
```
Dwire, line 196: Great job using the global scope to declare `numberInline`

index.js
```diff
@@ -1,81 +1,120 @@
...
-// console.log("%cQuestion 1", "color: red") 
-
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+ console.log("%cQuestion 1", "color: red") 
+const counter = 1
+console.log(counter) 
+// => 2
+ console.log("%c----------", "color: red")
```
Dwire, line 7: we want to be able to re-assign the counter variable which means that we should declare it with `let` instead of `const`. 

let counter = 1
counter = 2
console.log(counter)

Dwire (commented):
Great Job! a little slip up with `let` vs `const` in the first deliverable, and then you didn't get to the bonus, but outside of that your code looks very good.

PR #68, opened 2021-02-02
not ready for scope

evaldman:


index.js
```diff
@@ -1,88 +1,117 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+let counter = 1
+counter = 2
+console.log("%cQuestion 1", "color: red")
```

index.js
```diff
@@ -1,88 +1,117 @@
...
-
-// console.log(drinkWater(12))
+console.log("%cQuestion 3", "color: red") 
+function drinkWater(currentThirstLevel){
+  console.log("Man I sure am thirsty")
+  currentThirstLevel -= 1
+  console.log("Ahh that hits the spot")
+  return currentThirstLevel
+}
+// def drink_water(current_thirst_level)
```
BrendaMichelle, line 35: Good! 💯 One possible way to condense this and not lose any readability would be to consolidate lines 32 and 33 to:
```js
return currentThirstLevel - 1
```

index.js
```diff
@@ -1,88 +1,117 @@
...
+function sameSameButDifferent (num, maybeNum){
+  if (num === maybeNum){
+    return "same same"
+  } else if (num == maybeNum){
+    return "same same (but different)"
+  } else {
+    return "different"
+  }
+}
+console.log(sameSameButDifferent(5, 5))
```
BrendaMichelle, line 61: Nice! Another solution would be to convert the string to a number with parseInt to avoid `== `:

index.js
```diff
@@ -1,88 +1,117 @@
...
-// updateGrade(student1, 92)
-// console.log(student1)
+console.log("%cQuestion 5", "color: red")
+
+function updateGrade (student, grade){
+  student.grade = grade
+}
+const student1 = { name: "Duane", grade: 88 }
+updateGrade(student1, 92)
+console.log(student1)
```
BrendaMichelle, line 85: Good!

index.js
```diff
@@ -1,88 +1,117 @@
...
-//     name: "Liza", phones: { cell: "555-234-5678", office: "555-567-1234" }
-//   }
-// ]
-// printNameAndPhones(users)
+console.log("%cQuestion 6", "color: red")
+function printNameAndPhones(users){
+  users.forEach(element => console.log(element.name))
+  users.forEach(element => console.log(element.phones.cell))
+  users.forEach(element => console.log(element.phones.office))
+}
```
BrendaMichelle, line 98: Nice job using arrow functions! I see you're iterating over the array 3 times. The functionality required in this function to be done in one iteration, which is better for performance.

- Here's how it would look using an arrow function as the callback to `forEach` in one iteration:
```javascript
  users.forEach(user => {
    console.log(user.name);
  })
```

index.js
```diff
@@ -100,23 +129,27 @@ function myMap(array, callback) {
...
 
 // ***** Callbacks - Question 2 *****
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 2", "color: red")
-
-// console.log(myMap(["Raffy", "Chase"], greet)) 
+console.log("%cCallbacks - Question 2", "color: red")
+function greet (name){
+  return `Hi, ${name}!`
+}
```
BrendaMichelle, line 149: Nice!

BrendaMichelle (commented):
Well done! I see you didn't get to the last few ones, which of course is totally ok! Let us know if you have any questions about them or come to office hours so we can help you work through them 🙂 If you re-attempt them, also feel free to re-submit and send us a message on Slack, so we can review your updates

PR #69, opened 2021-02-22
Mini challenge solution

celestialheathen:


index.js
```diff
@@ -2,6 +2,9 @@
 // *** Uncomment the lines below to test
 // console.log("%cQuestion 1", "color: red") 
 
+let counter = 1;
+counter = 2;
+
```
BrendaMichelle, line 7: Looks good!

index.js
```diff
@@ -37,6 +50,12 @@
 // *** Uncomment the lines below to test
 // console.log("%cQuestion 4", "color: red") 
 
+function sameSameButDifferent(num, maybeNum) {
+  if (num === maybeNum) { console.log ("same same")}
+  else if ( num == maybeNum) { console.log ("same same (but different)")}
+  else { console.log ("different")}
+}
```
BrendaMichelle, line 57: Nice! Another solution would be to convert the string to a number with parseInt to avoid `==` :

index.js
```diff
@@ -67,6 +92,14 @@
...
 // console.log("%cQuestion 6", "color: red")
 
+function printNameAndPhones(users) {
+  users.forEach(function(user_hash) {
+    console.log(user_hash["name"])
+    console.log(`Cell: ${user_hash["phones"]["cell"]}`)
+    console.log(`Office: ${user_hash["phones"]["office"]}`)
+  })
+}
+
```
BrendaMichelle, line 102: - Nice! Keep in mind it's convention to use [camelCase](https://en.wikipedia.org/wiki/Camel_case) in Javascript over snake case (`user_hash`) and also, in Javascript we refer to each user as an `object` instead of a `hash`. So instead of `user_hash`, a more Javascript-esque name would be `userObject`
- Here's how this solution would look using dot notation instead of bracket notation:
```javascript
    console.log(user.name);
  })
```

index.js
```diff
@@ -13,6 +16,8 @@
 // *** otherwise, the error will stop the rest of your code from running
 // console.log("%cQuestion 2", "color: red") 
 
+const NAME = "James Yu";
```

index.js
```diff
@@ -114,6 +151,10 @@ function myMap(array, callback) {
 // *** Uncomment the lines below to test
 // console.log("%cCallbacks - Question 2", "color: red")
 
+function greet(name) {
+  return `Hi, ${name}!`
+}
+
```
BrendaMichelle, line 157: 💯

index.js
```diff
@@ -139,6 +180,13 @@ const line = []
 
 // ***** Scope & Closures - Question 1 *****
 
+let ticketNum = 1
+
+function takeATicketNumber(line) {
+  line.push(ticketNum)
+  ticketNum++
+  return `Welcome. You are ticket number ${line.length}`
+}
```
BrendaMichelle, line 189: Good! How would the functionality and use case of this function differ if you returned `ticketNum` instead of `line.length`?

index.js
```diff
@@ -162,6 +210,18 @@ const line = []
...
+  let current_num = 1
+  let current_line = []
+
+  function takeATicketNumber(current_line) {
+    current_line.push(current_num)
+    console.log(`Welcome. You are ticket number ${current_line.length}`)
+    current_num++
+  }
+  return takeATicketNumber
+}
```
BrendaMichelle, line 224: - If `takeATicketNumber` accepts an array, does `current_line` on line 216 serve a purpose? Try taking 216 out & testing it out
- Better variable names for `current_num` and `current_line` are `currentNum` and `currentLine`

BrendaMichelle (commented):
Great work! Left some comments

PR #70, opened 2021-02-22
Max-Beneke-Solutions

MaxBeneke:
'Done'

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,88 +1,117 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+let counter = 1;
+counter = 2;
+console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+console.log(counter) 
+// => 2
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
-
-// console.log(name)
+console.log("%cQuestion 2", "color: red") 
+const name = "Max";
+console.log(name);
 // => "Raffy"
-// name = "Not Raffy" 
+// name = "Not Raffy"; 
 // => TypeError
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red") 
 
 
 
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
 
-// console.log(drinkWater(12))
-// // => "Man I sure am thirsty"
-// // => "Ahh that hits the spot"
-// // => 11
-// console.log("%c----------", "color: red") 
+function drinkWater(currentThirstLevel) {
+  console.log("Man I sure am thirsty");
+  currentThirstLevel -= 1;
+  console.log("Ahh that hits the spot");
+  return currentThirstLevel
+}
+
+console.log(drinkWater(12))
+// => "Man I sure am thirsty"
+// => "Ahh that hits the spot"
+// => 11
+console.log("%c----------", "color: red") 
 
 
 
 // ***** Question 4 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 4", "color: red") 
+console.log("%cQuestion 4", "color: red") 
 
-// console.log(sameSameButDifferent(5, 5)) 
-// // => "same same"
+function sameSameButDifferent(num, maybeNum) {
+  if (num === maybeNum) {
+    return "same same";
+  } else if (num == maybeNum) {
+    return "same same (but different)";
+  } else {
+    return "different";
+  }
+}
+console.log(sameSameButDifferent(5, 5)) 
+// => "same same"
 
-// console.log(sameSameButDifferent(123, "123")) 
-// // => "same same (but different)"
+console.log(sameSameButDifferent(123, "123")) 
+// => "same same (but different)"
 
-// console.log(sameSameButDifferent(5, 7)) 
-// // => "different"
+console.log(sameSameButDifferent(5, 7)) 
+// => "different"
 
-// console.log(sameSameButDifferent(123, "122")) 
-// // => "different"
-// console.log("%c----------", "color: red") 
+console.log(sameSameButDifferent(123, "122")) 
+// => "different"
+console.log("%c----------", "color: red") 
 
 
 
 // ***** Question 5 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 5", "color: red")
-// const student1 = { name: "Duane", grade: 88 }
-// updateGrade(student1, 92)
-// console.log(student1)
-// // => { name: "Duane", grade: 92 }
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,88 +1,117 @@
...
 
 
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
+
+function drinkWater(currentThirstLevel) {
+  console.log("Man I sure am thirsty");
+  currentThirstLevel -= 1;
```
emotionalproof, line 32: This is great. You can optionally do currentThirstLevel--

index.js
```diff
@@ -1,88 +1,117 @@
...
+
+function sameSameButDifferent(num, maybeNum) {
+  if (num === maybeNum) {
+    return "same same";
+  } else if (num == maybeNum) {
+    return "same same (but different)";
+  } else {
+    return "different";
+  }
+}
```
emotionalproof, line 57: This is perfect for the problem and 100% correct since the purpose was to demonstrate the difference between == and ===. If you wanted to dig a little deeper and confirm the datatype, you can use typeof and do 
  if (typeof num === "number" && typeof maybeNum === "number" && num === maybeNum) {
    return "same same"
  } else if (typeof num === "number" && typeof maybeNum == "number" && num == maybeNum) {
  } else {
    return "different"
  }
}

You could also avoid using '==' and do 

index.js
```diff
@@ -1,88 +1,117 @@
...
-// console.log("%cQuestion 5", "color: red")
-// const student1 = { name: "Duane", grade: 88 }
-// updateGrade(student1, 92)
-// console.log(student1)
-// // => { name: "Duane", grade: 92 }
-// console.log("%c----------", "color: red")
+console.log("%cQuestion 5", "color: red")
+
+function updateGrade(student, number) {
+  student.grade = number;
```
emotionalproof, line 78: nice job using dot notation

index.js
```diff
@@ -1,88 +1,117 @@
...
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
+
+function drinkWater(currentThirstLevel) {
+  console.log("Man I sure am thirsty");
+  currentThirstLevel -= 1;
+  console.log("Ahh that hits the spot");
+  return currentThirstLevel
+}
```
emotionalproof, line 35: This is great. You can optionally do currentThirstLevel--.on line 32

or you could combine lines 32 and 34 and do 
return currentThirstLevel - 1

index.js
```diff
@@ -1,88 +1,117 @@
...
-// // => "Duane"
+console.log("%cQuestion 6", "color: red")
+
+function printNameAndPhones(arr) {
+  arr.forEach(function callBack(value, i, arr) {
+    console.log(arr[i]["name"]);
+    console.log(`Cell: ${arr[i]["phones"]["cell"]}`);
+    console.log(`Office: ${arr[i]["phones"]["office"]}`);
+  })
+}
```
emotionalproof, line 98: Nice job making this work. We can simplify your code by using the built-in power of forEach. forEach is going to iterate over the array "users" below. We can use "user" to represent each instance in the array and then put that inside a callback function.
users.forEach(function callback(user) {}

Since callback only exists inside of printNameAndPhones and will never be called elsewhere, we can actually make "callback" into an anonymous function and remove its name
users.forEach(function(user) {}

Now we can rewrite the console.log 's without having to do arr[i].  You can use dot or bracket notation 

    console.log(user.name)
    console.log("Cell:", user['phones']['cell'])
    console.log("Office:", user.phones.office)


function printNameAndPhones(users) {
    console.log(user.name)
    console.log("Cell:", user['phones']['cell'])
    console.log("Office:", user.phones.office)
  })
}

The most important thing is that you were able to make it work. What I provided would be a good refactor.

index.js
```diff
@@ -100,23 +129,29 @@ function myMap(array, callback) {
...
 
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 1", "color: red")
+console.log("%cCallbacks - Question 1", "color: red")
 
-// console.log(myMap([1,2,3,4,5], triple)) 
-// // => [3,6,9,12,15]
+function triple(value, i, arr) {
+  return value * 3;
+}
```
emotionalproof, line 136: This works however, you are calling the default arguments for forEach (value, i, arr). Those would be the correct parameters if  the triple and greet functions were being called by forEach but they're being called by myMap, which is a custom function that only passes in one argument. the "i" and "arr" you added are essentially placeholders for extra arguments myMap could pass to triple when called.  Instead, you can just do 

function triple(value) {
   return value * 3
}

take a look at the 5th line of myMap in the README to see how myMap is calling the callback

index.js
```diff
@@ -100,23 +129,29 @@ function myMap(array, callback) {
...
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 2", "color: red")
+console.log("%cCallbacks - Question 2", "color: red")
 
-// console.log(myMap(["Raffy", "Chase"], greet)) 
-// // => ["Hi, Raffy!", "Hi, Chase!"]
-// console.log("%c----------", "color: red") 
+function greet(value, i, arr) {
+  return `Hi, ${value}!`;
+}
```
emotionalproof, line 151: see above

index.js
```diff
@@ -141,44 +176,66 @@ const line = []
...
+console.log("%cScope & Closures - Question 1", "color: red")
+let ticketNumber = 0;
 
-// console.log(takeATicketNumber(line))
-// // => `Welcome. You are ticket number 1`
+function takeATicketNumber(arr) {
+  ticketNumber++;
+  line.push(ticketNumber)
+  return `Welcome. You are ticket number ${ticketNumber}`
+}
```
emotionalproof, line 186: perfect

index.js
```diff
@@ -141,44 +176,66 @@ const line = []
...
+      arr.push(1);
+      counter.push(1);
+    } else {
+      arr.push(parseInt(`${arr.length + 1}`));
+      counter.push(parseInt(`${arr.length + 1}`));
+    }
+    return `Welcome. You are ticket number ${counter.length}`;
+  }
+  return newTakeATicketNumber
+}
```
emotionalproof, line 222: This function works but can be significantly cleaned up by using a clue from the readme. "Write a higher order function called ticketNumberGeneratorFunc that returns a nested takeATicketNumber function. The other functionality should remain the same."
In other words we just need to use the function from the previous question inside of the new ticketNumberGeneratorFunce and return it

  let ticket = 0

  return function takeATicketNumber(line) {
    ticket++
    line.push(ticket)
    return `Welcome, you are ticket number ${ticket}`
  }
}

index.js
```diff
@@ -141,44 +176,66 @@ const line = []
...
+// => `Currently serving 2.`
 
-// console.log(takeATicketNumberFunc(newLine))
-// // => `Welcome. You are ticket number 3`
+console.log(takeATicketNumberFunc(newLine))
+// => `Welcome. You are ticket number 3`
 
-// console.log("%c----------", "color: red") 
\ No newline at end of file
+console.log("%c----------", "color: red")
```
emotionalproof, line 241: Great job Max. Looks like you are picking up the foundations nicely. Make sure you are referencing MDN docs frequently. Even if you know the answer, you can dig a little deeper. This will help your understanding of JS and make it easier to refactor.

emotionalproof (commented):
nice work!

PR #71, opened 2021-02-22
Completed mini-challenge

maquino96:
Please advise on if I approached the Scope and Closure problems correctly. Thanks in advance!

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,88 +1,121 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+let counter = 1
+counter = 2
+
+console.log(counter) 
+// => 2
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red") 
 
-// console.log(name)
+const name = 'Matt'
+
+console.log(name)
 // => "Raffy"
 // name = "Not Raffy" 
 // => TypeError
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red") 
 
 
 
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
 
-// console.log(drinkWater(12))
+function drinkWater(thirstLevel){
+  console.log('Man I sure am thirsty');
+  thirstLevel--;
+  console.log('Ahh that hits the spot');
+  return thirstLevel
+}
+
+console.log(drinkWater(12))
 // // => "Man I sure am thirsty"
 // // => "Ahh that hits the spot"
 // // => 11
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red") 
 
 
 
 // ***** Question 4 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 4", "color: red") 
+console.log("%cQuestion 4", "color: red") 
 
-// console.log(sameSameButDifferent(5, 5)) 
+function sameSameButDifferent(num, maybeNum){
+  if (typeof maybeNum === 'number' && num === maybeNum){
+    return "same same"
+  } else if (num == maybeNum) {
+    return "same same (but different)";
+  } else { return "different"}
+
+}
+
+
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,88 +1,121 @@
...
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+let counter = 1
+counter = 2
+
```
BrendaMichelle, line 7: Looking good!

index.js
```diff
@@ -1,88 +1,121 @@
...
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red") 
+
+const name = 'Matt'
```

index.js
```diff
@@ -1,88 +1,121 @@
...
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
 
-// console.log(drinkWater(12))
+function drinkWater(thirstLevel){
+  console.log('Man I sure am thirsty');
+  thirstLevel--;
+  console.log('Ahh that hits the spot');
+  return thirstLevel
+}
```
BrendaMichelle, line 38: 💯

index.js
```diff
@@ -1,88 +1,121 @@
...
+
+function sameSameButDifferent(num, maybeNum){
+  if (typeof maybeNum === 'number' && num === maybeNum){
+    return "same same"
+  } else if (num == maybeNum) {
+    return "same same (but different)";
+  } else { return "different"}
 
-// console.log(sameSameButDifferent(5, 5)) 
+}
```
BrendaMichelle, line 59: Nice! Another solution would be to convert the string to a number with parseInt to avoid == :

index.js
```diff
@@ -1,88 +1,121 @@
...
-//   { 
-//     name: "Liza", phones: { cell: "555-234-5678", office: "555-567-1234" }
-//   }
-// ]
-// printNameAndPhones(users)
+console.log("%cQuestion 6", "color: red")
+
+function printNameAndPhones(users){
+  users.forEach( el => console.log(el.name,'\n',"Cell: "+el.phones.cell,'\n',"Office: "+el.phones.office))
+}
```
BrendaMichelle, line 100: Nice! Using a [template string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) may look cleaner in this case since there's no need for the newline character or the addition operator. Example:

```suggestion
function printNameAndPhones(users){
  users.forEach( el => {
  console.log(`${el.name}
  Cell: ${el.phones.cell}
  Office: ${el.phones.office}`)
    )
  }
}
```

index.js
```diff
@@ -100,25 +133,34 @@ function myMap(array, callback) {
...
 
-// console.log(myMap(["Raffy", "Chase"], greet)) 
-// // => ["Hi, Raffy!", "Hi, Chase!"]
-// console.log("%c----------", "color: red") 
+console.log("%cCallbacks - Question 2", "color: red")
 
 
+function greet(name){
+  return `Hi, ${name}!`
+}
```
BrendaMichelle, line 158: 💯

index.js
```diff
@@ -136,49 +178,75 @@ function nowServing(line) {
...
 // *** Uncomment the lines below to test
-// console.log("%cScope & Closures - Question 1", "color: red")
+console.log("%cScope & Closures - Question 1", "color: red")
+
+function takeATicketNumber(line){
+  let nextNum = extraLine.length+1
+  line.push(nextNum)
+  extraLine.push(nextNum)
+  return `Welcome, You are ticket number ${nextNum}`
+}
```
BrendaMichelle, line 194: You are using `extraLine` mostly for its length property, which you increase by 1 in order to get a value for `nextNum`. Then, you are pushing the `nextNum` into the `extraLine` array solely to increase the length of the array.  While this may work, this can be simplified. What is the purpose of storing nextNum in this array (especially if we are never actually using these stored values from the `extraLine` array)? Think about replacing `extraLine` with a variable that holds a number data type instead.

index.js
```diff
@@ -136,49 +178,75 @@ function nowServing(line) {
...
+    line.push(nextNum)
+    extraLine.push(nextNum)
+    return `Welcome, You are ticket number ${nextNum}`
+  }
+
+  return takeATicketNumber2
+}
+
+const newLine = [] // goals is NOT to use a global variable to store data
+const takeATicketNumberFunc = ticketNumberGeneratorFunc()
```
BrendaMichelle, line 233: Consider the above comment as well for this one in relation to `extraLine`. You can still use the global variable `newLine` on line 233 (the directions were hinting at a solution for this problem while hinting at the solution for the previous problem).

BrendaMichelle (commented):
Good work! Left some comments

PR #74, opened 2021-02-22
Done

DayneDewar:


index.js
```diff
@@ -1,88 +1,127 @@
...
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red") 
+const name = "Dayne";
```
emotionalproof, line 20: nice use of let vs const

index.js
```diff
@@ -1,88 +1,127 @@
...
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
+
+function drinkWater(currentThirstLevel) {
+  console.log("Man I sure am Thirsty")
+  currentThirstLevel -= 1;
+  console.log("Ahh that hits the spot");
+  return currentThirstLevel;
+}
```
emotionalproof, line 39: This is great. We can also refactor line 36 as currentThirstLevel--
Or better yet, merge lines36 and 38 and just do 
return currentThirstLevel - 1

index.js
```diff
@@ -1,88 +1,127 @@
...
+  if (num === maybeNum) {
+      return "same same";
+  }
+  else if (num == maybeNum) {
+    return "same same (but different)";
+  } 
+  else {
+    return "different";
+  }
+}
```
emotionalproof, line 63: Nice! The key was to demonstrate the different between == and ===. Another option for the else if would be to use parseInt to convert maybeNum to an integer
else if (num === parseInt(maybeNum))

If you wanted to dive a little deeper and confirm the data type, you could use typeof, e.g.:
if (typeof num === 'number' && typeof maybeNum === 'number && num === maybeNum)

index.js
```diff
@@ -1,88 +1,127 @@
...
-// const student1 = { name: "Duane", grade: 88 }
-// updateGrade(student1, 92)
-// console.log(student1)
-// // => { name: "Duane", grade: 92 }
-// console.log("%c----------", "color: red")
+console.log("%cQuestion 5", "color: red")
+const student1 = { name: "Duane", grade: 88 }
+
+function updateGrade(student, grade) {
+  student.grade = grade;
```
emotionalproof, line 86: nice dot notation

index.js
```diff
@@ -1,88 +1,127 @@
...
+function printNameAndPhones(users) {
+  users.forEach(iterate)
+}
+
+function iterate(hash) {
+  console.log(hash.name);
+  console.log(`Cell: ${hash.phones.cell}`);
+  console.log(`Office: ${hash.phones.office}`);
+}
+
```
emotionalproof, line 108: Great! This works with iterate as an external function that gets referenced. Practice this with iterate as an anonymous callback in the forEach.

index.js
```diff
@@ -100,23 +139,29 @@ function myMap(array, callback) {
 // ***** Callbacks - Question 1 *****
 
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 1", "color: red")
+console.log("%cCallbacks - Question 1", "color: red")
 
-// console.log(myMap([1,2,3,4,5], triple)) 
-// // => [3,6,9,12,15]
+function triple(num) {
+  return (num * 3)
```
emotionalproof, line 145: This is perfect but just as a note, you can omit the ()

index.js
```diff
@@ -141,44 +186,68 @@ const line = []
...
+  let num = 0;
+  function takeATicketNumber(arr){
+    num++;
+    arr.push;
+
+    return `Welcome. You are ticket number ${num}`
+  }
+
+  return takeATicketNumber
+}
```
emotionalproof, line 236: This is great. we can refactor by removing line 235 and adding the return in front of line 228. We can also make the function on 228 an anonymous function if you want.

emotionalproof (commented):
Great work Dayne. Some notes on refactoring mostly. Make sure to check the mdn docs frequently to deepen your understanding

PR #75, opened 2021-02-22
done ?

jamesdesousa:


index.js
```diff
@@ -1,88 +1,119 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
-
-// console.log(counter) 
+console.log("%cQuestion 1", "color: red")
+let counter = 1
+counter = 2
+console.log(counter)
```
BrendaMichelle, line 6: Good!  `let` allows us to reassign our variables

index.js
```diff
@@ -1,88 +1,119 @@
...
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
-// *** after testing, comment the line causing the error back in 
+// *** after testing, comment the line causing the error back in
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red")
+const name = "James"
```
BrendaMichelle, line 16: Nice! The key here was using the keyword `const` because it doesn't allow us to reassign our variable

index.js
```diff
@@ -1,88 +1,119 @@
...
+      return ("same same")
+    } else {
+      return ("same same (but different)")
+    }
+  }
+  if (num != parseInt(maybeNum)) {
+      return ("different")
+  }
+}
+console.log(sameSameButDifferent(5, 5))
```
BrendaMichelle, line 58: This works! Check out this [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness) about equality comparison.

- Your first if-statement is checking if `num` is loosely equivalent to `maybeNum` ([`==`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Equality)). You could get away with not using `parseInt` to convert `maybeNum` in that conditional.
- Check out another possible solution using the [strict equality operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators#identity). This solution assumes `num` is always a Number data type.
```js
  if (num === maybeNum) {
    return 'same same'
  }
    return 'same same (but different)'
  }
  else {
    return 'different'
  }
}
```

- Check out this solution where we don't assume `num` is a Number data type. The only change is in the first conditional statement
```js
  if (typeof num === 'number' && num === maybeNum) {
    return 'same same'
  }
    return 'same same (but different)'
  }
  else {
    return 'different'
  }
}
```

index.js
```diff
@@ -1,88 +1,119 @@
...
+
+function updateGrade(student, grade) {
+  student.grade = grade
+}
+
+
+
+const student1 = { name: "Duane", grade: 88 }
+updateGrade(student1, 92)
+console.log(student1)
```
BrendaMichelle, line 84: Good 👍 The updates made to the `student` object inside the function is reflected in the `student1` object because objects are ["call by sharing" unlike primitive data types](https://stackoverflow.com/questions/518000/is-javascript-a-pass-by-reference-or-pass-by-value-language).

index.js
```diff
@@ -1,88 +1,119 @@
...
+
+const users = [
+  {
+    name: "Duane", phones: { cell: "555-123-4567", office: "555-456-7890" }
+  },
+  {
+    name: "Liza", phones: { cell: "555-234-5678", office: "555-567-1234" }
+  }
+]
+printNameAndPhones(users)
```
BrendaMichelle, line 109: Wonderful 👏

- Here's how the body of the function for this solution would look with for..of:
```javascript
  for (let user of users) {
    console.log(user.name);
  }
```

- Here's how the body of the function for this solution would look with forEach:
```javascript
    console.log(user.name);
  })

File: index.js [hunk 6]
```diff
@@ -100,23 +131,32 @@ function myMap(array, callback) {
...
 
 // ***** Callbacks - Question 2 *****
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 2", "color: red")
+console.log("%cCallbacks - Question 2", "color: red")
 
-// console.log(myMap(["Raffy", "Chase"], greet)) 
+function greet(name) {
+  return `Hi ${name}!`
+}
```
BrendaMichelle, line 154: Good!

index.js
```diff
@@ -138,32 +178,38 @@ function nowServing(line) {
 const line = []
 
 // ***** Scope & Closures - Question 1 *****
-
+let i = 0
+function takeATicketNumber(line) {
+   i += 1
+   line.push(i)
+   return `Welcome. You are ticket nummber ${i}`
+}
```
BrendaMichelle, line 186: Very good! The key here was using a global variable to save the ticket number

index.js
```diff
@@ -138,32 +178,38 @@ function nowServing(line) {
...
 
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red")
 
 // ***** Scope & Closures - Question 2 *****
 // *** Uncomment the lines below to test
-// console.log("%cScope & Closures - Question 2", "color: red")
+console.log("%cScope & Closures - Question 2", "color: red")
 
+console.log("Going to finish working on the last one tonight")
```
BrendaMichelle, line 212: Were you able to solve this?
- Check out [Ian's scrimba on closures](https://scrimba.com/scrim/cLgEmZUZ)
- Or the end of the optional [functions lecture from yesterday](https://youtu.be/RQW3ydDJodA). Jump to `28:30` for the closures section
- Let me know if you have any questions or want to work through this together

BrendaMichelle (commented):
Good work! Left some comments.

PR #77, opened 2021-02-22
My first pull request

amuesing:
This is my first pull request to get feedback from mini challenge. Is this the correct way to do it?

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,54 +1,77 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+let counter = 1
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+console.log("%cQuestion 1", "color: red") 
+
+console.log(counter)
+console.log(counter+1)
+// => 2
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red") 
 
-// console.log(name)
+const name = "Aaron"
+
+console.log(name)
 // => "Raffy"
 // name = "Not Raffy" 
 // => TypeError
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red") 
 
 
 
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,88 +1,129 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+let counter = 1
```
BrendaMichelle, line 3: Nice 💯  `let` allows us to reassign our variables

index.js
```diff
@@ -1,88 +1,129 @@
...
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red") 
+
+const name = "Aaron"
```

index.js
```diff
@@ -1,88 +1,129 @@
...
+  if (num === maybeNum) {
+    return(`same smae`)
+  }
+  if (num == maybeNum) {
+    return(`same same (but different)`)
+  }
+  if (num != maybeNum) {
+    return(`different`)
+  }
+}
```

index.js
```diff
@@ -1,88 +1,129 @@
...
-// console.log(student1)
-// // => { name: "Duane", grade: 92 }
-// console.log("%c----------", "color: red")
+console.log("%cQuestion 5", "color: red")
+
+const student1 = { name: "Duane", grade: 88 }
+
+function updateGrade(student, grade) {
+student.grade = grade
+}
```
BrendaMichelle, line 86:
```suggestion
function updateGrade(student, grade) {
	student.grade = grade
}
```

index.js
```diff
@@ -1,88 +1,129 @@
...
+//     console.log(arr[i]);
 //   }
-// ]
-// printNameAndPhones(users)
-// // => "Duane"
+// }
+
+function printNameAndPhones(arr) {
+arr.forEach(element => console.log(element));
+}
```
BrendaMichelle, line 118: Nice use of arrow functions! This problem wanted you to console log specific values from each object. Check out this solution:

```javascript
users.forEach(function (user) {
    console.log(user.name);
})
```

index.js
```diff
@@ -1,88 +1,129 @@
...
-// console.log(drinkWater(12))
-// // => "Man I sure am thirsty"
-// // => "Ahh that hits the spot"
-// // => 11
-// console.log("%c----------", "color: red") 
+function drinkWater(currentThirstLevel) {
+ console.log("Man I sure am thirsty");
+ console.log("Ahh that hits the spot");
+ return(currentThirstLevel-1)
+}
```
BrendaMichelle, line 37: Good!

PR #78, opened 2021-02-22
Done

jahaira-marrero:


index.js
```diff
@@ -1,88 +1,119 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+var counter = 1
```
emotionalproof, line 3: This works but since ES6, let and const are preferred for declaring variables.  As a rule of thumb in this course, never use var.  If you use let, JS will let you mutate or change the variable to something else. 
let dog = "maizey"
dog.toUpperCase()  (mutating the string to upper case)
dog = "Chase" (saving a different string to the same variable)
dog = [1,2,3,4] (saving an array to the same variable)

If you use const, you can mutate what is saved but you cannot change it
dog.toUpperCase()  (mutating the string to upper case. mutation is allowed with const)
dog = "Chase" (this would cause an error)

Mutation is taking the existing item and changing it a little such as adding or removing an item from array

Once you use const to declare a variable, you cannot use "=" with it again or you will get an error. Try playing around with this in the console in your browser.

index.js
```diff
@@ -1,88 +1,119 @@
...
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red") 
+const name = "Jahaira"
+console.log(name)
```
emotionalproof, line 18: perfect

index.js
```diff
@@ -1,88 +1,119 @@
...
-// console.log("%c----------", "color: red") 
+// let currentThirstLevel = 5; 
+
+function drinkWater(currentThirstLevel) {
+  if (currentThirstLevel < 15) {
+    return "Man I sure an thirsty"}
+    else {
+    return " Ahh that hits the spot"
+  }
+}
```
emotionalproof, line 38: Nice work on structuring this function but this does not create the desired output. The goal is to have
// => "Man I sure am thirsty"
// => "Ahh that hits the spot"
// => 11
All three of the above lines print and our source is the ruby method.
def drink_water(current_thirst_level)
  puts "Man I sure am thirsty"
  current_thirst_level -= 1
  puts "Ahh that hits the spot"
  current_thirst_level
end

First thing we can do is setup the envelope for the function which you did perfectly

function drinkWater(currentThirstLevel) {
}

Next we need to account for the "puts' in ruby.  Remember that puts is just printing something to the console, it is not a return value. The equivalent to puts in js is console.log(). So we can do:

Next we need to account for "current_thirst_level -= 1" and decrement or lower the value of that variable by one. This line will be nearly identical in JS except instead of using snake_case like we do in Ruby, we are going to use camelCase
 currentThirstLevel -= 1

Next we will handle the second puts

Finally, we will return the currentThirstLevel.In  ruby, the last line of the method will automatically be returned. This is not the case in JS so we need to use the word 'return'

return currentThirstLevel

Putting it altogether:
function drinkWater(currentThirstLevel) {
  currentThirstLevel-= 1
  return currentThirstLevel
}

index.js
```diff
@@ -1,88 +1,119 @@
...
+console.log("%cQuestion 4", "color: red") 
+
+function sameSameButDifferent(num, maybeNum) {
+  if (num === maybeNum) {
+    return  "same same";
+  } else if (num == maybeNum) {
+    return "same same (but different)";
+  } else { return "different"}
+}
+console.log(sameSameButDifferent(5, 5))
```
emotionalproof, line 58: This is perfect. The key was to demonstrate the difference between == and ===. Now that you have done that, try refactoring the else if using parseInt.

To go deeper, use typeof in the conditional to determine the data type
e.g.

index.js
```diff
@@ -1,88 +1,119 @@
...
-// const student1 = { name: "Duane", grade: 88 }
-// updateGrade(student1, 92)
-// console.log(student1)
-// // => { name: "Duane", grade: 92 }
-// console.log("%c----------", "color: red")
+console.log("%cQuestion 5", "color: red")
+const student1 = { name: "Duane", grade: 88 }
+
+function updateGrade(student, grade) {
+  student.grade = grade
```
emotionalproof, line 79: nice use of dot notation

index.js
```diff
@@ -1,88 +1,119 @@
...
+
+function printNameAndPhones(users) {
+  for (let i = 0; i < users.length; i++){
+    console.log(users[i].name);
+    console.log(`cell: ${users[i].phones.cell}`);
+    console.log(`office: ${users[i].phones.office}`);
+  }
+}
+printNameAndPhones(users)
+
```
emotionalproof, line 109: Great job! Now try refactoring using forEach and for ... of

index.js
```diff
@@ -100,23 +131,29 @@ function myMap(array, callback) {
...
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 2", "color: red")
+console.log("%cCallbacks - Question 2", "color: red")
 
-// console.log(myMap(["Raffy", "Chase"], greet)) 
-// // => ["Hi, Raffy!", "Hi, Chase!"]
-// console.log("%c----------", "color: red") 
+function greet(name) {
+  return `Hi, ${name}`
+}
```
emotionalproof, line 153: nice string interpolation!

index.js
```diff
@@ -132,53 +169,58 @@ function nowServing(line) {
...
+console.log("%cScope & Closures - Question 1", "color: red")
 
-// console.log(takeATicketNumber(line))
-// // => `Welcome. You are ticket number 1`
+function takeATicketNumber(line) {
+  line.push(ticketNumber)
+  ticketNumber += 1
+  return `Welcome. You are ticket number ${ticketNumber - 1}.`
+}
+console.log(takeATicketNumber(line))
```
emotionalproof, line 188: Nice job!  How could you make this work if you started ticketNumber at 0?

Swap lines 184 and 185. Then remove the -1 from like 186

index.js
```diff
@@ -132,53 +169,58 @@ function nowServing(line) {
...
-// console.log("%cScope & Closures - Question 2", "color: red")
+console.log("%cScope & Closures - Question 2", "color: red")
 
-// const newLine = []
-// const takeATicketNumberFunc = ticketNumberGeneratorFunc()
-// console.log(takeATicketNumberFunc(newLine))
-// // => `Welcome. You are ticket number 1`
+const newLine = []
+const takeATicketNumberFunc = ticketNumberGeneratorFunc()
+console.log(takeATicketNumberFunc(newLine))
```
emotionalproof, line 211: Try giving this a shot and let me know if you want to discuss.

emotionalproof (commented):
Great work Jahaira. With the exception of the use of var and translating ruby to js, it looks like you are taking to the fundamentals nicely.  Make sure you look at the MDN docs frequently to deepen your understanding

PR #79, opened 2021-02-22
First pull request

ajperlis07:


index.js
```diff
@@ -1,88 +1,114 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+console.log("%cQuestion 1", "color: red") 
+counter = 1
+counter = 2
```
BrendaMichelle, line 6: While this works, it's best practice to declare variables with `let` or `const`. In this case, we would need to use `let` because `let` allows us to reassign our variables

index.js
```diff
@@ -1,88 +1,114 @@
...
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
-
-// console.log(name)
+console.log("%cQuestion 2", "color: red") 
+const name = "Andrew Perlis"
```
BrendaMichelle, line 17: Great! The key here was using `const` because it doesn't allow us to reassign our variable

index.js
```diff
@@ -1,88 +1,114 @@
...
-// // => "Man I sure am thirsty"
-// // => "Ahh that hits the spot"
-// // => 11
-// console.log("%c----------", "color: red") 
+console.log("%cQuestion 3", "color: red") 
+function drinkWater(currentThirstLevel) {
+  console.log("Man I sure am thirsty")
+  currentThirstLevel -= 1
+  console.log("Ahh that hits the spot")
+  return currentThirstLevel
```
BrendaMichelle, line 33: Good!

index.js
```diff
@@ -1,88 +1,114 @@
...
-// console.log(sameSameButDifferent(5, 5)) 
+console.log("%cQuestion 4", "color: red") 
+function sameSameButDifferent(num, maybeNum){
+  if (num === maybeNum){
+  return "same same"
+} else if (num == maybeNum){
+  return "same same (but different)"
+} else 
+    return "different"
+}
```
BrendaMichelle, line 53: Very good! Another solution would be to convert the string to a number with parseInt to avoid `==` :

index.js
```diff
@@ -100,23 +126,29 @@ function myMap(array, callback) {
 // ***** Callbacks - Question 1 *****
 
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 1", "color: red")
+function triple(arrayValue) {
+  return arrayValue * 3;
+}
```
BrendaMichelle, line 131: 💯

index.js
```diff
@@ -100,23 +126,29 @@ function myMap(array, callback) {
...
+// => [6,12,18,24]
+console.log("%c----------", "color: red") 
 
 
 // ***** Callbacks - Question 2 *****
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 2", "color: red")
+function greet(name){
+  return "Hi, " + name + "!"
+}
```
BrendaMichelle, line 146: Good! Another solution would be to use template strings to string interpolate:

```js
function greet(name){
  return `Hi, ${name}!`
}
```

index.js
```diff
@@ -141,44 +173,64 @@ const line = []
...
 
-// console.log(takeATicketNumber(line))
-// // => `Welcome. You are ticket number 1`
+  var ticketNumber = 1
+  function takeATicketNumber(line) {
+    line.push(ticketNumber)
+    var newMsg = "Welcome. You are ticket number " + ticketNumber
+      ticketNumber++;
+      return newMsg;
+  }
```
BrendaMichelle, line 183: Good! One suggestion is to stick with `let` and `const` for variable declarations. Since ES6, this has been the best practice because it helps prevent some scoping issues related with using `var`. Check out[ this video ](https://www.youtube.com/watch?v=wtbZCWnZIyQ&feature=emb_title)to learn more about why

index.js
```diff
@@ -141,44 +173,64 @@ const line = []
...
+//   var message = "Welcome. You are ticket number " + ticketNumberNew
+//   ticketNumberNew++;
+//   return message
+// }
+console.log("%cScope & Closures - Question 2", "color: red")
+
+const newLine = []
+const takeATicketNumberFunc = function ticketNumberGeneratorFunc(newLine){
+  return takeATicketNumber(newLine);
+}
```
BrendaMichelle, line 220: Were you able to figure this out since? Let me know if you want to work through this together!

BrendaMichelle (commented):
Great work! Left some comments

PR #80, opened 2021-02-23
Almost Done

kgyapong1:


index.js
```diff
@@ -3,35 +3,43 @@
...
 
 // console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+// => 2
+console.log("%c----------", "color: red") 
 
+console.log("%cQuestion 1", "color: red") 
+counter = 1
+counter = 2
```
emotionalproof, line 11: In order to declare a variable we must use let, const, or var. For the purposes of this course we are going to ignore var and just focus on let and const.  let is used when we want to have the ability to save a new value to the variable later so the answer to this problem is 
let counter = 1

index.js
```diff
@@ -3,35 +3,43 @@
...
+console.log(counter)
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red") 
+const name = "Kwaku"
```
emotionalproof, line 19: nice use of const!

index.js
```diff
@@ -3,35 +3,43 @@
...
-// console.log("%c----------", "color: red") 
 
+console.log("%cQuestion 3", "color: red") 
 
+function drinkWater(thirst) {
+  console.log("Man I sure am thristy")
+  thirst -= 1
+  console.log("Ahh that hits the spot")
+  return thirst
+}
```
emotionalproof, line 40: this is great! on line 37 you could also do thirst--
you could also combine lines 37 and 39 and just do
return thirst - 1

index.js
```diff
@@ -50,7 +58,21 @@
...
+  if (num === maybeNum) {
+    console.log("same same")
+  }
+  else if (num == maybeNum) {
+    console.log("same same (but different)")
+  }
+  else {
+    console.log("different")
+  }
+}
```
emotionalproof, line 71: perfect.  Another option for line 65 would be to use parseInt to convert maybeNum to an integer and then use ===

if you want to dig a little deeper you can use typeof to determine the data type
e.g.

index.js
```diff
@@ -50,7 +58,21 @@
...
+console.log(sameSameButDifferent(5, 5))
```
emotionalproof, line 72: Good job on the completed deliverables. Please at least make an attempt on all questions for these assessments so we can give feedback.

emotionalproof (commented):
Nice work Kwaku but please make an attempt on all deliverables and resubmit

PR #81, opened 2021-02-23
Done

ZacharyDagnall:


index.js
```diff
@@ -1,88 +1,131 @@
...
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
-
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+console.log("%cQuestion 1", "color: red")
+let counter = 1
+counter = 2
+console.log(counter)
```

index.js
```diff
@@ -1,88 +1,131 @@
...
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
-
-// console.log(name)
+console.log("%cQuestion 2", "color: red")
+const name = "Zak"
```
BrendaMichelle, line 16: Good! The key here was using `const` because it doesn't allow us to reassign our variable

index.js
```diff
@@ -1,88 +1,131 @@
...
+//   puts "Ahh that hits the spot"
+//   current_thirst_level
+// end
+
+function drinkWater(currentThirstLevel) {
+  console.log("Man I sure am thirsty")
+  currentThirstLevel -= 1
+  console.log("Add that hits the spot")
+  return currentThirstLevel
+}
```
BrendaMichelle, line 41: Good!

index.js
```diff
@@ -1,88 +1,131 @@
...
-// console.log("%cQuestion 4", "color: red") 
+console.log("%cQuestion 4", "color: red")
+function sameSameButDifferent(num, maybeNum) {
+  if (num === maybeNum) {
+    return "same same"
+  } else if (num == maybeNum) {
+    return "same same (but different)"
+  } else {
+    return "different"
+  }
```

index.js
```diff
@@ -98,87 +141,91 @@ function myMap(array, callback) {
...
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 2", "color: red")
-
-// console.log(myMap(["Raffy", "Chase"], greet)) 
-// // => ["Hi, Raffy!", "Hi, Chase!"]
-// console.log("%c----------", "color: red") 
+console.log("%cCallbacks - Question 2", "color: red")
+function greet(name) {
+  return `Hi, ${name}!`
+}
```
BrendaMichelle, line 163: 💯

index.js
```diff
@@ -98,87 +141,91 @@ function myMap(array, callback) {
...
-// console.log("%c----------", "color: red") 
+console.log("%cScope & Closures - Question 1", "color: red")
+function ticketNumberMaker(line) {
+  nextNum = 1
+  return function ticketMaker(line) {
+    line.push(nextNum)
+    nextNum++
+    return `Welcome. You are ticket number ${nextNum - 1}`
+  }
+}
```
BrendaMichelle, line 194: Looks like this the solution for Question 2; looks good! One thing you could do to simplify the logic is initialize `nextNum` to 0 instead of 1 on line 188. Like so:
```suggestion
function ticketNumberMaker(line) {
  nextNum = 0
  return function ticketMaker(line) {
    nextNum++
    line.push(nextNum)
    return `Welcome. You are ticket number ${nextNum}`
  }
}
```

BrendaMichelle (commented):
Good work! Left some comments

PR #82, opened 2021-02-23
JS Mini Challenge 1 submission

mzthomas961:
Done.

### DIFF
--- a/index.js
+++ b/index.js
@@ -157,7 +157,7 @@ function greet(name) {
 function takeANumber(line, name) {
   line.push(name)
 
-  return `Welcome, ${name}. You are number ${line.length} in line.`
+  return `Welcome, ${name}. You're number ${line.length} in line.`
 }
 
 function nowServing(line) {
--- a/index.js
+++ b/index.js
@@ -1,7 +1,8 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
 // console.log("%cQuestion 1", "color: red") 
-
+// var counter = 1;
+// counter = 2;
 // console.log(counter) 
 // // => 2
 // console.log("%c----------", "color: red") 
@@ -12,11 +13,11 @@
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
 // console.log("%cQuestion 2", "color: red") 
-
+//  const name = "Matthew Thomas";
 // console.log(name)
-// => "Raffy"
+// // => "Raffy"
 // name = "Not Raffy" 
-// => TypeError
+// // => TypeError
 // console.log("%c----------", "color: red") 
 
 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,7 +1,8 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
 // console.log("%cQuestion 1", "color: red") 
-
+// var counter = 1;
+// counter = 2;
```
emotionalproof, line 5: This works but for the purpose of this course, we will just focusing on let and const.  Since ES6, let and const are preferred since there are some scoping issues with var.  The answer we were looking for for this was let counter = 1

index.js
```diff
@@ -12,19 +13,24 @@
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
 // console.log("%cQuestion 2", "color: red") 
-
+//  const name = "Matthew Thomas";
```
emotionalproof, line 16: Perfect

index.js
```diff
@@ -12,19 +13,24 @@
...
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
 // console.log("%cQuestion 3", "color: red") 
-
+function drinkWater(currentThirstLevel){
+console.log("Man I sure am thirsty")
+currentThirstLevel -= 1
+console.log ("Ahh that hits the spot")
+return currentThirstLevel;
+}
```
emotionalproof, line 33: great! line 30 could also be currentThirstLevel--
you could also combine lines 30 and 32 and just do 
return currentThirstLevel - 1

index.js
```diff
@@ -40,7 +46,22 @@
...
+  else if (num == maybeNum)
+  {
+    return "same same but different"
+  }
+  else
+  {
+    return "Different."
+  }
+  }
+// console.log(sameSameButDifferent(123, '123'))
```
emotionalproof, line 64: Perfect. You could also try playing around with parseInt in the else if and avoid having to use ==

you could also dig deeper and use typeof to check the data type of the variable

index.js
```diff
@@ -51,6 +72,9 @@
 // console.log("%c----------", "color: red") 
 
 
+function updateGrade(student, grade) {
+  student.grade = grade
```
emotionalproof, line 76: nice use of dot notation

index.js
```diff
@@ -112,6 +140,10 @@ function myMap(array, callback) {
 
 // ***** Callbacks - Question 2 *****
 // *** Uncomment the lines below to test
+
+function greet(name) {
+  return `Hi ${name}`
+}
```
emotionalproof, line 146: nice string interpolation

index.js
```diff
@@ -139,26 +171,31 @@ const line = []
...
 
+let ticket = 0
 
-// *** Uncomment the lines below to test
-// console.log("%cScope & Closures - Question 1", "color: red")
+function takeATicketNumber(line) {
+  ticket += 1
+  line.push(ticket)
+  return `Welcome. You're ticket number ${ticket}`
+}
```
emotionalproof, line 180: this is great

index.js
```diff
@@ -139,26 +171,31 @@ const line = []
...
+console.log(nowServing(line))
+// => `Currently serving 2.`
 
-// console.log(takeATicketNumber(line))
-// // => `Welcome. You are ticket number 3`
+console.log(takeATicketNumber(line))
+// => `Welcome. You are ticket number 3`
 
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red")
```
emotionalproof, line 198: Give the last deliverable a shot

emotionalproof (commented):
Nice job! Make sure you reviewing the MDN docs frequently to deepen your understanding of the material

PR #83, opened 2021-02-23
done

pat-tusza:


index.js
```diff
@@ -1,4 +1,7 @@
 // ***** Question 1 *****
+
+let counter = 1
+counter = 2
```
BrendaMichelle, line 4: Nice 💯  `let` allows us to reassign our variables

index.js
```diff
@@ -8,6 +11,10 @@
 
 
 // ***** Question 2 *****
+
+const name = "Patrick"
```
BrendaMichelle, line 15: Great! The key here was using `const` because it doesn't allow us to reassign our variable

index.js
```diff
@@ -34,6 +49,18 @@
...
+  else if (num == maybeNum) {
+    return "same same (but different)"
+  }
+  else {
+    return "different"
+  }
+  
+}
 // *** Uncomment the lines below to test
 // console.log("%cQuestion 4", "color: red")
```
BrendaMichelle, line 65: Nice! Another solution would be to convert the string to a number with parseInt to avoid `==` :

index.js
```diff
@@ -111,6 +151,9 @@ function myMap(array, callback) {
 
 
 // ***** Callbacks - Question 2 *****
+function greet(name) {
+  return `Hi, ${name}!`
+}
```
BrendaMichelle, line 156: Good!

index.js
```diff
@@ -139,7 +182,13 @@ const line = []
 
 // ***** Scope & Closures - Question 1 *****
 
+let ticNum = 0
 
+function takeATicketNumber(line){
+  ticNum += 1
+  line.push(ticNum)
+  return `Welcome. you are ticket number ${ticNum}`
+}
```
BrendaMichelle, line 191: 💯

index.js
```diff
@@ -161,6 +210,16 @@ const line = []
...
 // ***** Scope & Closures - Question 2 *****
+
+function ticketNumberGeneratorFunc(){
+  let num = 0
+  return function (line) {
+    num += 1
+    line.push(num)
+    return `Welcome. you are ticket number ${num}`
+  }
+}
```
BrendaMichelle, line 221: Nice!

BrendaMichelle (commented):
Great work! Keep it up!

PR #84, opened 2021-02-23
almost done

mikeansong7:


index.js
```diff
@@ -6,32 +6,45 @@
...
+
+
 // console.log(name)
 // => "Raffy"
 // name = "Not Raffy" 
 // => TypeError
 // console.log("%c----------", "color: red") 
 
+const name = `michael`
+
```
emotionalproof, line 29: nice use of let and const!

index.js
```diff
@@ -6,32 +6,45 @@
...
+// => "Ahh that hits the spot"
+// => 11
+console.log("%c----------", "color: red") 
+
+function drinkWater(currentThirstLevel) {
+  console.log("Man I sure am thirsty");
+  currentThirstLevel -= 1;
+  console.log("Ahh that hits the spot");
+  return currentThirstLevel
+}
```
emotionalproof, line 47: perfect! you can all do currentThirsLevel-- 
Or combine lines 44 and 46 and just do return currentThirstLevel - 1

index.js
```diff
@@ -50,6 +63,16 @@
...
+function sameSameButDifferent(num, maybeNum) {
+  if(num === maybeNum){
+  return "same same";
+} elsif (num == maybeNum){
+      return "same same (but different)";
+    } else{
+    return "different"
+  }
+}
+
```
emotionalproof, line 75: This is great. Please keep an eye on your indentation so your code stays organized.
You can try playing around with parseInt to avoid using the ==

To go deeper you can use typeof to check the datatype of the variables as well

emotionalproof (commented):
Nice work Mike! But I would like to see you try to do all of the questions with these assessments so we can provide feedback.

PR #86, opened 2021-03-15
Da-mell's mini challenge

DamellMorningstar:


index.js
```diff
@@ -1,92 +1,99 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
-
-// console.log(counter) 
+console.log("%cQuestion 1", "color: red")
+let counter = 1;
```
lindsayMecher, line 4: awesome! 'let' allows you to reassign the variable later.

index.js
```diff
@@ -1,92 +1,99 @@
...
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
-// *** after testing, comment the line causing the error back in 
+// *** after testing, comment the line causing the error back in
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
-
-// console.log(name)
+console.log("%cQuestion 2", "color: red")
+const name = "Da-mell"
```
lindsayMecher, line 14: great job! 'const' is for when you do not want to be able to reassign your variable.

index.js
```diff
@@ -1,92 +1,99 @@
...
-
-
+console.log("%c----------", "color: red")
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
-
-// console.log(drinkWater(12))
+console.log("%cQuestion 3", "color: red")
+const drinkWater = (currentThirstLevel) => {
```
lindsayMecher, line 23: nice use of arrow functions!! Fun fact - if there is only one parameter, then you don't need the paranthesis! In all other cases you do need them.

`const drinkWater = currentThirstLevel => {`

index.js
```diff
@@ -1,92 +1,99 @@
...
+  if (currentThirstLevel >= 12)
+  {
+    console.log("Man I sure am thirsty")
+     currentThirstLevel -= 1
+  }
+  console.log("Ahh that hits the spot")
+  return currentThirstLevel
+}
```
lindsayMecher, line 31: nice! I don't think you needed the constraint for only if the currentThirstLevel was >= 12 but it's a cool idea! Another way you can write line 27 is 

`currentThirstLevel --`

index.js
```diff
@@ -1,92 +1,99 @@
...
+console.log("%c----------", "color: red")
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
-
-// console.log(drinkWater(12))
+console.log("%cQuestion 3", "color: red")
+const drinkWater = (currentThirstLevel) => {
+  if (currentThirstLevel >= 12)
+  {
```
lindsayMecher, line 25: this curly brace syntax is widely debated. It really depends on the language, in C#, this is the correct syntax, in JS you will more commonly see

```
if (currentThirstLevel >= 12) {
 // code block
}
```

index.js
```diff
@@ -1,92 +1,99 @@
...
-
-
+console.log("%c----------", "color: red")
 // ***** Question 4 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 4", "color: red") 
-
-// console.log(sameSameButDifferent(5, 5)) 
+console.log("%cQuestion 4", "color: red")
+const sameSameButDifferent = (num, maybenum) => {
```
lindsayMecher, line 40: I would say to write this the following way to follow the camelCase convention.

`maybeNum`

index.js
```diff
@@ -1,92 +1,99 @@
...
+  if (num === maybenum) {
+    return "same same"
+  } else if (num == maybenum) {
+    return "same same(but different)"
```
lindsayMecher, line 44: line 45 should be 

`} else {`

Basically line 46 gets moved up to be on the same line as 45.

index.js
lindsayMecher, line 40: I like this solution! Nice job :)

index.js
```diff
@@ -1,92 +1,99 @@
...
-// const student1 = { name: "Duane", grade: 88 }
-// updateGrade(student1, 92)
-// console.log(student1)
+console.log("%cQuestion 5", "color: red")
+const updateGrade = (student, grade) => {
+  return student.grade = grade
+}
+const student1 = { name: "Duane", grade: 88 }
+updateGrade(student1, 92)
+console.log(student1)
```
lindsayMecher, line 67: nice!

index.js
```diff
@@ -1,92 +1,99 @@
...
+    name: "Liza", phones: { cell: "555-234-5678", office: "555-567-1234" }
+  }
+]
+const printNameAndPhones = (user) => {
+  for (i = 0; i < users.length; i++){
+    console.log(users[i]["name"])
+    console.log(users[i]["phones"]["cell"])
+    console.log(users[i]["phones"]["office"])
+  }
+}
```
lindsayMecher, line 87: Da-mell, nice work! You are just missing the string interpolation. Here is a simplified version.

```
const printNameAndPhones = usersArr => {
  usersArr.forEach(userObject => {
    console.log(userObject.name);
    console.log(`Cell: ${userObject.phones.cell}`);
    console.log(`Office: ${userObject.phones.office}`);
  })
}
```

index.js
```diff
@@ -1,92 +1,99 @@
...
 // => "Liza"
 // => "Cell: 555-234-5678"
 // => "Office: 555-567-1234"
-// console.log("%c----------", "color: red") 
-
-
+// console.log("%c----------", "color: red")
 // ***** Callbacks *****
-
 function myMap(array, callback) {
```
lindsayMecher, line 97: Da-Mell, you were on the way to the solution. Here is my solution, first we create the function triple.

```
const triple = num => {
  return num * 3;
}
```
which is the same thing you have, but it is not expecting an array as an argument, but a single number.

This can be further simplified down to a one-liner.

`const triple = num => num * 3;`

as for the function myMap, we want it to map over an existing array and then alter each element in the array using the callback function

```
const myMap = (arr, fn) => {
  return arr.map(item => {
    return fn(item);
  })
}
```

we're saying, hey, return a new array, which is going to use the array we pass in, and on each element in that array, we will invoke the callback function. We need to say return on both lines, because we want to return each individual value in the array, and then return the whole new array. This can be further simplified to either:

```
const myMap = (arr, fn) => {
  return arr.map(item => fn(item));
}
```

or 

`const myMap = (arr, fn) => arr.map(item => fn(item));`

index.js
```diff
@@ -96,89 +103,72 @@ function myMap(array, callback) {
...
 // ***** Callbacks - Question 2 *****
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 2", "color: red")
-
-// console.log(myMap(["Raffy", "Chase"], greet)) 
+console.log("%cCallbacks - Question 2", "color: red")
+const greet = (arr) => {
+  return `Heyyy, ${arr}`
+}
+console.log(myMap(["Raffy", "Chase"], greet))
```
lindsayMecher, line 123: Here is my solution, first define greet, which you have,

```
const greet = name => {
  return `Hi, ${name}!`;
}
```

or simplify to 

`const greet = name => `Hi, ${name}!`;`

then define the myMap function

```
const myMap = (arr, fn) => {
  return arr.map(name => {
    return fn(name);
  })
}
```

or simplified

`const myMap = (arr, fn) => arr.map(name => fn(name));`

we are creating a new array using map, where each element is altered by applying the callback function.

index.js
```diff
@@ -96,89 +103,72 @@ function myMap(array, callback) {
...
 // *** Uncomment the lines below to test
 // console.log("%cScope & Closures - Question 1", "color: red")
-
-// console.log(takeATicketNumber(line))
+// function takeTickerNumber(line) {
+//   ticket += 1;
+//   line.push(ticket)
+//   return `Welcome. You are ticket number ${ticket}`
+// }
+// console.log(takeTickerNumber(line))
```
lindsayMecher, line 146: Nice work, you just have to change one thing, take the ticket variable out of the function into the global scope. Otherwise, it cannot be called again and updated. It will always be 1.

```
let ticketNumber = 0;

const takeATicketNumber = line => {
  ticketNumber += 1;
  line.push(ticketNumber);
  return `Welcome. You are ticket number ${ticketNumber}`;
}
```

index.js
```diff
@@ -96,89 +103,72 @@ function myMap(array, callback) {
...
-
 // console.log(nowServing(newLine))
 // // => `Currently serving 1.`
-
 // console.log(nowServing(newLine))
 // // => `Currently serving 2.`
-
 // console.log(takeATicketNumberFunc(newLine))
 // // => `Welcome. You are ticket number 3`
+// console.log("%c----------", "color: red")
```
lindsayMecher, line 172: This could be solved in the following way

`const ticketNumberGeneratorFunc = () => {
  let number = 0;

  return (line) => {
    number += 1;
    line.push(number);
    return `Welcome. You are ticket number ${number}`;
  }
}`

It is basically just encapsulating all the logic from the previous question into one function.

lindsayMecher (commented):
Great work, Da-Mell!! Impressive use of arrow functions.

PR #99, opened 2021-03-16
Feedback

bleakchandler:
Hi can you give me feedback? Thanks!

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,89 +1,136 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+
+
+console.log("%cQuestion 1", "color: red") 
+
+var counter = 1
+counter = 2
+
+console.log(counter) 
+// => 2
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
 
-// console.log(name)
+
+console.log("%cQuestion 2", "color: red") 
+
+// const name = "Bleak"
+
+console.log(name)
 // => "Raffy"
-// name = "Not Raffy" 
+name = "Not Raffy" 
 // => TypeError
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red") 
 
 
 
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
 
-// console.log(drinkWater(12))
-// // => "Man I sure am thirsty"
-// // => "Ahh that hits the spot"
-// // => 11
-// console.log("%c----------", "color: red") 
+function drinkWater(num) {
+  console.log("Man I sure am thirsty")
+  num -= 1
+  console.log("Ahh that hits the spot")
+  return num
+}
 
+console.log(drinkWater(12))
+// => "Man I sure am thirsty"
+// => "Ahh that hits the spot"
+// => 11
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 4 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 4", "color: red") 
+console.log("%cQuestion 4", "color: red") 
 
-// console.log(sameSameButDifferent(5, 5)) 
-// // => "same same"
 
-// console.log(sameSameButDifferent(123, "123")) 
-// // => "same same (but different)"
+function sameSameButDifferent (num, maybeNum) {
+  if (num === maybeNum) {
+//=== checks to see if the two values are equal values and equal types
+  console.log("same same") 
+  }
+  else if (num == maybeNum) {
+    console.log("same same (but different)") 
+  }
+  else {
+    console.log("different") 
+  }
+  }
 
-// console.log(sameSameButDifferent(5, 7)) 
-// // => "different"
 
-// console.log(sameSameButDifferent(123, "122")) 
-// // => "different"
-// console.log("%c----------", "color: red") 
+console.log(sameSameButDifferent(5, 5)) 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,89 +1,136 @@
...
-// console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+
+
+console.log("%cQuestion 1", "color: red") 
+
+var counter = 1
```
lindsayMecher, line 8: You will want to avoid variable declarations with `var`. It will only be used in legacy code moving forward because it can cause problems relating to scope. Here would be a good place to use `let`

index.js
```diff
@@ -1,89 +1,136 @@
...
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
 
-// console.log(name)
+
+console.log("%cQuestion 2", "color: red") 
+
+// const name = "Bleak"
```
lindsayMecher, line 24: awesome! `const` will not allow you to reassign the variable.

index.js
```diff
@@ -1,89 +1,136 @@
...
-// // => "Man I sure am thirsty"
-// // => "Ahh that hits the spot"
-// // => 11
-// console.log("%c----------", "color: red") 
+function drinkWater(num) {
+  console.log("Man I sure am thirsty")
+  num -= 1
+  console.log("Ahh that hits the spot")
+  return num
+}
```
lindsayMecher, line 43: great! you can also further simplify this:

```
function drinkWater(thirstLevel) {
  console.log("Man I sure am thirsty");
  console.log("Ahh that hits the spot");
  return thirstLevel -= 1;
}
```

index.js
```diff
@@ -1,89 +1,136 @@
...
+//=== checks to see if the two values are equal values and equal types
+  console.log("same same") 
+  }
+  else if (num == maybeNum) {
+    console.log("same same (but different)") 
+  }
+  else {
+    console.log("different") 
+  }
+  }
```
lindsayMecher, line 68: great! Just be mindful of the if/else statement conventions.

```
  if (num === maybeNum) {
    console.log("same same")
  } else if (num == maybeNum) {
    console.log("same same but different")
  } else {
    console.log("different")
  }
}
```

index.js
```diff
@@ -1,89 +1,136 @@
...
-// const student1 = { name: "Duane", grade: 88 }
-// updateGrade(student1, 92)
-// console.log(student1)
-// // => { name: "Duane", grade: 92 }
-// console.log("%c----------", "color: red")
+console.log("%cQuestion 5", "color: red")
+
+function updateGrade (student, grade) {
+
+  student.grade = grade
```
lindsayMecher, line 92: great!

index.js
```diff
@@ -1,89 +1,136 @@
...
+]
+
+function printNameAndPhones(users) {
+
+  for (let i = 0; i < users.length; ++i) {
+        console.log(users[i].name);
+        console.log("Cell:", users[i].phones.cell);
+        console.log("Office:", users[i].phones.office);
+  }
+}
```
lindsayMecher, line 124: great!! You can also use the forEach method to further simplify

```
function printNameAndPhones(usersArr) {
  usersArr.forEach(function(userObject) {
    console.log(userObject.name);
    console.log(`Cell: ${userObject.phones.cell}`)
    console.log(`Office: ${userObject.phones.office}`)
  })
}
```

index.js
```diff
@@ -100,25 +147,33 @@ function myMap(array, callback) {
 // ***** Callbacks - Question 1 *****
 
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 1", "color: red")
+console.log("%cCallbacks - Question 1", "color: red")
+
+function triple(value) {
+  var newValue = value * 3
+  return newValue
+}
```
lindsayMecher, line 155: great! You can further simplify to

```
function triple(num){
  return num * 3;
}
```

index.js
```diff
@@ -100,25 +147,33 @@ function myMap(array, callback) {
...
-
-// console.log(myMap(["Raffy", "Chase"], greet)) 
-// // => ["Hi, Raffy!", "Hi, Chase!"]
-// console.log("%c----------", "color: red") 
+console.log("%cCallbacks - Question 2", "color: red")
 
+function greet(name) {
+  greeting = (`Hi, ${name}!`);
+  return greeting;
+}
```
lindsayMecher, line 172: Great! Can be further simplified:

```
function greet(name) {
  return `Hi, ${name}!`;
}
```

index.js
```diff
@@ -136,36 +191,49 @@ function nowServing(line) {
...
+function takeATicketNumber (line) {
 
-// *** Uncomment the lines below to test
-// console.log("%cScope & Closures - Question 1", "color: red")
+  ticketNum += 1
+  return `Welcome. You are ticket number ${ticketNum}.`
 
-// console.log(takeATicketNumber(line))
-// // => `Welcome. You are ticket number 1`
+}
```
lindsayMecher, line 205: good. Would use `let` on line 194.

also you are missing pushing the ticketNumber into the line

```
let ticketNumber = 0

function takeATicketNumber(line) {
  ticketNumber += 1
  line.push(ticketNumber)
  return `Welcome. You are ticket number ${ticketNumber}`
}
```

lindsayMecher: this way the line variable is getting updated so we can keep track of all the tickets.

index.js
```diff
@@ -136,36 +191,49 @@ function nowServing(line) {
...
-// console.log("%cScope & Closures - Question 2", "color: red")
+console.log("%cScope & Closures - Question 2", "color: red")
+
+const newLine = []
+const takeATicketNumberFunc = ticketNumberGeneratorFunc()
+
+function ticketNumberGeneratorFunc(newLine) {
+
+  
+}
```
lindsayMecher, line 234: Can be solved by:

```
  let number = 0;
  return function (line) {
    number += 1;
    line.push(number);
    return `Welcome. You are ticket number ${number}`;
  }
}
```

Basically all our logic is encapsulated in this function so we can pass in the argument of any line and then run all the logic of the first deliverable

lindsayMecher (commented):
Great work, Bleak! :)

PR #101, opened 2021-04-05
Done

MelSwell:
I really don't understand closures!!! 

BTW.. this is submission for Seth Cohen... just in case my github name is misleading

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,88 +1,121 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+console.log("%cQuestion 1", "color: red") 
+let counter = 1;
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+counter++;
+
+console.log(counter); 
+ // => 2
+ console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
-
-// console.log(name)
+console.log("%cQuestion 2", "color: red") 
+const name = "Seth";
+console.log(name);
 // => "Raffy"
 // name = "Not Raffy" 
 // => TypeError
-// console.log("%c----------", "color: red") 
-
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red")
 
-// console.log(drinkWater(12))
+const drinkWater = currentThirstLevel => {
+  console.log("Man I sure am thirsty");
+  currentThirstLevel -= 1;
+  console.log("Ahh that hits the spot")
+  return currentThirstLevel;
+}
+
+
+console.log(drinkWater(12))
 // // => "Man I sure am thirsty"
 // // => "Ahh that hits the spot"
 // // => 11
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red") 
 
 
 
 // ***** Question 4 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 4", "color: red") 
+console.log("%cQuestion 4", "color: red") 
 
-// console.log(sameSameButDifferent(5, 5)) 
+const sameSameButDifferent = (arg1, arg2) => {
+  if (arg1 === arg2){
+    return "same same"
+  } else if (arg1 == arg2){
+    return "same same (but different)"
+  } else {
+    return "different"
+  }
+}
+
+console.log(sameSameButDifferent(5, 5)) 
 // // => "same same"
 
-// console.log(sameSameButDifferent(123, "123")) 
+console.log(sameSameButDifferent(123, "123")) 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,88 +1,121 @@
...
-
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red")
+
+const drinkWater = currentThirstLevel => {
```
seanpadden, line 30: Nice use of arrow functions! Totally not required now but good to get comfortable with them sooner than later.

index.js
```diff
@@ -1,88 +1,121 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+console.log("%cQuestion 1", "color: red") 
+let counter = 1;
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+counter++;
```
seanpadden, line 6: This does what we want in that it changes the value to 2 (the first time around), and proper use of `let` vs `const` but really all you had to do was change the value:  `counter = 2` !

index.js
```diff
@@ -1,88 +1,121 @@
...
-// updateGrade(student1, 92)
-// console.log(student1)
+console.log("%cQuestion 5", "color: red")
+const student1 = { name: "Duane", grade: 88 }
+
+const updateGrade = (student, newGrade) => {
+  student.grade = newGrade;
+  return student;
+}
+
```
seanpadden, line 84: Good :+1: The updates made to the `student` object inside the function is reflected in the `student1` object because objects are ["call by sharing" unlike primitive data types](https://stackoverflow.com/questions/518000/is-javascript-a-pass-by-reference-or-pass-by-value-language).

index.js
```diff
@@ -1,88 +1,121 @@
...
+const sameSameButDifferent = (arg1, arg2) => {
+  if (arg1 === arg2){
+    return "same same"
+  } else if (arg1 == arg2){
+    return "same same (but different)"
+  } else {
+    return "different"
+  }
+}
```
seanpadden, line 59: Nice! The key here was understanding the difference between `===` and `==`. You could also convert the string to a number with parseInt to avoid == :

index.js
```diff
@@ -1,88 +1,121 @@
...
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
-
-// console.log(name)
+console.log("%cQuestion 2", "color: red") 
+const name = "Seth";
```
seanpadden, line 18: Great! The key here was using `const` because it doesn't allow us to reassign our variable

index.js
```diff
@@ -1,88 +1,121 @@
...
+    name: "Liza", phones: { cell: "555-234-5678", office: "555-567-1234" }
+  }
+]
+
+const printNameAndPhones = obj => {
+  for (const key in obj){
+    console.log(obj[key]);
+  }
+}
+
```
seanpadden, line 110: Nice! This works, but since this deliverable was specifically asking to print the users name and phone numbers, I would be a little more explicit in calling those specific attributes in your logs. And because `users` is an array, it might actually make more sense to use a `forEach` or `for...of` to iterate over the array. Your `for...in` works just fine, but is typically reserved for objects. Alternate implementations could look something like this:

```javascript
 for (let user of users) {
  console.log(user.name);
 }
```
```javascript
  console.log(user.name);
 })
```

index.js
```diff
@@ -100,23 +133,31 @@ function myMap(array, callback) {
 // ***** Callbacks - Question 1 *****
```
seanpadden, line 133: Looks like you missed this one? Here's an example of what creating our own `map` would look like though!

```javascript
function myMap(array, callback) {
  const result = []
  for (let i = 0; i < array.length; i++) {
    const element = array[i]
    const value = callback(element)
    result.push(value)
  }
  return result
}
```

index.js
```diff
@@ -136,49 +177,69 @@ function nowServing(line) {
...
+let ticketNumber = 0
 // ***** Scope & Closures - Question 1 *****
 
 
 // *** Uncomment the lines below to test
-// console.log("%cScope & Closures - Question 1", "color: red")
+console.log("%cScope & Closures - Question 1", "color: red")
+
+const takeATicketNumber = line => {
+  line.push(line.length + 1)
```
seanpadden, line 188: Cool. You could also push the `ticketNumber`

index.js
```diff
@@ -136,49 +177,69 @@ function nowServing(line) {
...
+console.log("%c----------", "color: red") 
 
 // ***** Scope & Closures - Question 2 *****
 // *** Uncomment the lines below to test
-// console.log("%cScope & Closures - Question 2", "color: red")
+console.log("%cScope & Closures - Question 2", "color: red")
+
+//This isn't working and I frankly don't even understand how it's even able to get
+// as far as it does lol... 
+
```
seanpadden, line 218: This one's a tricky bonus but you are so close!! Check this Scrimba out https://scrimba.com/scrim/cLgEmZUZ and see if you can figure this out after watching it and playing around with the examples. (In Scrimba, you can pause the video and play around with the code shown on the screen)

seanpadden:
Great job Seth! See my comments for some general feedback and tips.

PR #102, opened 2021-04-05
Done

m-lawrence:


index.js
```diff
@@ -1,88 +1,126 @@
 // ***** Question 1 *****
-// *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+let counter = 1
+counter += 1
+
```
seanpadden, line 5: Nice :100: `let` allows us to reassign our variables

You can also just change `counter`'s value by doing `counter = 2`

index.js
```diff
@@ -1,88 +1,126 @@
...
+// *** Uncomment the lines below to test
+console.log("%cQuestion 1", "color: red") 
 
+console.log(counter) 
+// => 2
+console.log("%c----------", "color: red") 
 
+const name = "Missy"
```
seanpadden, line 13: Great! The key here was using `const` because it doesn't allow us to reassign our variable

index.js
```diff
@@ -1,88 +1,126 @@
...
 // ***** Question 3 *****
+
+
+function drinkWater(currentThirstLevel) {
+  console.log("Man I sure am thirsty")
+  currentThirstLevel -=1 
+  console.log("Ahh that hits the spot")
+  return currentThirstLevel
+}
+
```
seanpadden, line 37: Good! One way to condense this and not lose any readability would be to consolidate lines 32 and 33 to:
```js
return currentThirstLevel - 1
```

```javascript
```
```javascript
```

index.js
```diff
@@ -1,88 +1,126 @@
...
+function sameSameButDifferent(num, maybeNum) {
+  if (num === maybeNum) {
+    return "same same"
+  } else if (num == maybeNum) {
+      return "same same (but different)"
+  } else {
+      return "different"
+  }
+}
```
seanpadden, line 59: Nice! The key here was understanding the difference between `===` and `==`. You could also convert the string to a number with parseInt to avoid == :

index.js
```diff
@@ -1,88 +1,126 @@
...
+console.log("%c----------", "color: red") 
 
 
 
 // ***** Question 5 *****
+
+function updateGrade(obj, num) {
+  obj["grade"] = num
+}
+
```
seanpadden, line 83: Nice! You can also use dot notation like so:
`obj.grade = num`

The updates made to the `student` object inside the function is reflected in the `student1` object because objects are ["call by sharing" unlike primitive data types](https://stackoverflow.com/questions/518000/is-javascript-a-pass-by-reference-or-pass-by-value-language). 


***I would also recommend using more clear variable names. Instead of `obj` and `num` you could call the arguments `student` and `grade`!

index.js
```diff
@@ -1,88 +1,126 @@
...
 // ***** Question 6 *****
+
+function printNameAndPhones(users) {
+  for (let i = 0; i < users.length; i++) {
+    console.log(users[i]["name"])
+    console.log(`Cell: ${users[i]["phones"]["cell"]}`)
+    console.log(`Office: ${users[i]["phones"]["office"]}`)
+  }
+}
+
```
seanpadden, line 103: Nice 👍  This for loop works, but for iterating over arrays, Javascript has a couple other methods that make it a little easier.

```javascript
 for (let user of users) {
  console.log(user.name);
 }
```
```javascript
  console.log(user.name);
 })
```

index.js
```diff
@@ -138,47 +183,63 @@ function nowServing(line) {
...
 // ***** Scope & Closures - Question 1 *****
+let ticketNumber = 0
 
+function takeATicketNumber(line) {
+  ticketNumber += 1
+  line.push(ticketNumber)
 
+  return `Welcome. You are ticket number ${ticketNumber}`
+}
 // *** Uncomment the lines below to test
```
seanpadden, line 194: Very nice! Great use of the global scope

index.js
```diff
@@ -138,47 +183,63 @@ function nowServing(line) {
...
 // ***** Scope & Closures - Question 2 *****
+
+function ticketNumberGeneratorFunc(newLine, takeATicketNumber) {
+    let ticketNumber = 0
+    ticketNumber += 1
+    newLine.push(ticketNumber)
+
+  return `Welcome. You are ticket number ${ticketNumber}`
+}
 // *** Uncomment the lines below to test
```
seanpadden, line 225: This one's a tricky bonus one! Check this Scrimba out https://scrimba.com/scrim/cLgEmZUZ and see if you can figure this out after watching it and playing around with the examples. (In Scrimba, you can pause the video and play around with the code shown on the screen)

PR #103, opened 2021-04-05
Mini Challenge Day 1

DianaLiao:
Hello! This is my first pull request ever! I thought I had solved the last question for the longest time but it turned out I was just reading my output wrong and it was super not solved >_< ...I think I finally got it in the end. Needed about 1hr 45min.

### DIFF
--- a/index.js
+++ b/index.js
@@ -43,20 +43,26 @@ console.log("%c----------", "color: red")
 
 // ***** Question 4 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 4", "color: red") 
+console.log("%cQuestion 4", "color: red") 
 
-// console.log(sameSameButDifferent(5, 5)) 
-// // => "same same"
+const sameSameButDifferent = (num, maybeNum) => {
+  if (num === maybeNum) {return "same same"}
+  else if (num == maybeNum) {return "same same (but different)"}
+  else if (num != maybeNum || num !== maybeNum) {return "different"}
+}
 
-// console.log(sameSameButDifferent(123, "123")) 
-// // => "same same (but different)"
+console.log(sameSameButDifferent(5, 5)) 
+// => "same same"
 
-// console.log(sameSameButDifferent(5, 7)) 
-// // => "different"
+console.log(sameSameButDifferent(123, "123")) 
+// => "same same (but different)"
 
-// console.log(sameSameButDifferent(123, "122")) 
-// // => "different"
-// console.log("%c----------", "color: red") 
+console.log(sameSameButDifferent(5, 7)) 
+// => "different"
+
+console.log(sameSameButDifferent(123, "122")) 
+// => "different"
+console.log("%c----------", "color: red") 
 
 
 
--- a/index.js
+++ b/index.js
@@ -128,14 +128,16 @@ function myMap(array, callback) {
 // ***** Callbacks - Question 1 *****
 
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 1", "color: red")
+console.log("%cCallbacks - Question 1", "color: red")
 
-// console.log(myMap([1,2,3,4,5], triple)) 
-// // => [3,6,9,12,15]
+const triple = num => num*3
 
-// console.log(myMap([2,4,6,8], triple))   
-// // => [6,12,18,24]
-// console.log("%c----------", "color: red") 
+console.log(myMap([1,2,3,4,5], triple)) 
+// => [3,6,9,12,15]
+
+console.log(myMap([2,4,6,8], triple))   
+// => [6,12,18,24]
+console.log("%c----------", "color: red") 
 
 
 // ***** Callbacks - Question 2 *****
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,88 +1,130 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
-
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+console.log("%cQuestion 1", "color: red") 
+let counter = 1
```
seanpadden, line 4: Nice :100: `let` allows us to reassign our variables

index.js
```diff
@@ -1,88 +1,130 @@
...
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
-
-// console.log(name)
+console.log("%cQuestion 2", "color: red") 
+const name = "Diana"
```

index.js
```diff
@@ -1,88 +1,130 @@
...
+console.log("%c----------", "color: red") 
 
 
 
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
+
+const drinkWater = currentThirstLevel => {
```
seanpadden, line 29: V nice~! Great use of arrow functions, which are not. at all mandatory atm but we'll be using more and more especially once we hit React. 

One way to condense this and not lose any readability would be to consolidate lines 31 and 33 to:
```js
return currentThirstLevel - 1
```

```javascript
```
```javascript
```

index.js
```diff
@@ -1,88 +1,130 @@
...
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 4", "color: red") 
+console.log("%cQuestion 4", "color: red") 
+
+
+const sameSameButDifferent = (num, maybeNum) => {
+  if (num === maybeNum) {return "same same"}
+  else if (num == maybeNum) {return "same same (but different)"}
+  else {return "different"}
+}
```
seanpadden, line 54: Nice! The key here was understanding the difference between `===` and `==`. You could also convert the string to a number with parseInt to avoid `==`

PS love the single line `if` return blocks - definitely no need to make a new line for a one line return! If you're using a one liner, you also don't need the `{ }` brackets, btw!

index.js
```diff
@@ -1,88 +1,130 @@
...
 
-// console.log(sameSameButDifferent(5, 5)) 
-// // => "same same"
+// * The solution below more literally fits the directions  *
```
seanpadden, line 55: ✅

index.js
```diff
@@ -1,88 +1,130 @@
...
-// // => { name: "Duane", grade: 92 }
-// console.log("%c----------", "color: red")
+console.log("%cQuestion 5", "color: red")
+const student1 = { name: "Duane", grade: 88 }
+
+const updateGrade = (student, grade) => {
+  student.grade = grade
+  return student
+}
+
```
seanpadden, line 92: Good :+1: The updates made to the `student` object inside the function is reflected in the `student1` object because objects are ["call by sharing" unlike primitive data types](https://stackoverflow.com/questions/518000/is-javascript-a-pass-by-reference-or-pass-by-value-language).

index.js
```diff
@@ -1,88 +1,130 @@
...
+]
+
+const printNameAndPhones = peeps => {
+  for (const user of peeps){
+    console.log(user.name)
+    console.log(`Cell: ${user.phones.cell}`)
+    console.log(`Office: ${user.phones.office}`)
+  }
+}
+printNameAndPhones(users)
```
seanpadden, line 120: Great! Both are equally good options, but here's how it would look with forEach:
```javascript
  console.log(user.name);
 })
```

index.js
```diff
@@ -141,44 +187,63 @@ const line = []
...
 
-// console.log(takeATicketNumber(line))
-// // => `Welcome. You are ticket number 2`
+let ticket = 0
+const takeATicketNumber = lineArray => {
+  ticket ++ 
+  lineArray.push(ticket)
+  return `Welcome. You are ticket number ${ticket}`
+}
+
```
seanpadden, line 199: v nice 💯 . great use of the global scope!

index.js
```diff
@@ -141,44 +187,63 @@ const line = []
...
+const ticketNumberGeneratorFunc = () => {
+  let ticketNumber = 0
+  const takeATicketNumber = (lineArray) => {
+    ticketNumber ++
+    lineArray.push(ticketNumber)
+    return `Welcome. You are ticket number ${ticketNumber}`
+  }
+  return takeATicketNumber
+}
+
```
seanpadden, line 230: Great!!! You can also return the function on line 223 likeeee this:

```javascript
const ticketNumberGeneratorFunc = () => {
  let ticketNumber = 0
  return function takeATicketNumber(lineArray) {
    ticketNumber ++
    lineArray.push(ticketNumber)
    return `Welcome. You are ticket number ${ticketNumber}`
  }
}
```

PR #104, opened 2021-04-06
minichallenge-1

chrisgk47:


index.js
```diff
@@ -1,42 +1,60 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
-
-// console.log(counter) 
+console.log("%cQuestion 1", "color: red") 
+let counter = 1;
```
seanpadden, line 4: Nice :100: `let` allows us to reassign our variables

index.js
```diff
@@ -1,42 +1,60 @@
...
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
-
+console.log("%cQuestion 2", "color: red") 
+const name = "Chris" //when using const, cant change/overwrite name. use const when you know for a fact
```

index.js
```diff
@@ -1,42 +1,60 @@
...
 
 
 
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
 
-// console.log(drinkWater(12))
+function drinkWater(currentThirstLevel){ //use camelCase
```
seanpadden, line 31: Great! One way to condense this and not lose any readability would be to consolidate lines 33 and 35 to:
```js
return currentThirstLevel - 1
```

```javascript
```
```javascript
```

index.js
```diff
@@ -1,42 +1,60 @@
...
 
 
 
 // ***** Question 4 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 4", "color: red") 
-
+console.log("%cQuestion 4", "color: red") 
+
+function sameSameButDifferent(num, maybeNum){
```
seanpadden, line 49: Nice! The key here was understanding the difference between `===` and `==`. Like you mention in the comments, you could also convert the string to a number with parseInt to avoid `==` :

index.js
```diff
@@ -54,28 +72,44 @@
...
 // ***** Question 5 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 5", "color: red")
-// const student1 = { name: "Duane", grade: 88 }
-// updateGrade(student1, 92)
+console.log("%cQuestion 5", "color: red")
+const student1 = { name: "Duane", grade: 88 }
+
+function updateGrade(student, grade){
+  student.grade = grade
```
seanpadden, line 79: Good :+1: The updates made to the `student` object inside the function is reflected in the `student1` object because objects are ["call by sharing" unlike primitive data types](https://stackoverflow.com/questions/518000/is-javascript-a-pass-by-reference-or-pass-by-value-language).

index.js
```diff
@@ -54,28 +72,44 @@
...
+
+const users = [ 
+  { 
+    name: "Duane", phones: { cell: "555-123-4567", office: "555-456-7890" }
+  },
+  { 
+    name: "Liza", phones: { cell: "555-234-5678", office: "555-567-1234" }
+  }
+]
+function print_name_and_phones(users){
```
seanpadden, line 104: Nice. callback function as an argument to `forEach` :white_check_mark: template strings for string interpolation :white_check_mark:
Btw, nice work naming the anonymous function's parameter something semantic!

- Also, Here's how it would look with for..of (both are equally good choices!):
```javascript
 for (let user of users) {
  console.log(user.name);
 }
```

index.js
```diff
@@ -142,7 +183,12 @@ const line = []
 
 // *** Uncomment the lines below to test
 // console.log("%cScope & Closures - Question 1", "color: red")
+let ticketNumber = 1
 
+function takeATicketNumber(){
```
seanpadden, line 188: Cool! Great use of the global scope. You could also start `ticketNumber` at 0 and increment before returning their line number. Same difference.

index.js
```diff
@@ -164,6 +210,19 @@ const line = []
 // *** Uncomment the lines below to test
 // console.log("%cScope & Closures - Question 2", "color: red")
 
+
+function ticketNumberGeneratorFunc(){
```
seanpadden, line 214: Very nice! Great use of closures. You can also actually return the function on line 218 to condense this even more:

```javascript
 let ticketNum = 0
  return function takeATicketNumber(newLine){
    ticketNum += 1
    line.push(ticketNum)
    return `Welcome, You are ticket number ${ticketNum}`
  }
}
```

PR #105, opened 2021-04-06
Done

HannaBM12:


index.js
```diff
@@ -1,88 +1,121 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
-
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+console.log("%cQuestion 1", "color: red") 
+let counter = 1;
+counter = 2
```
seanpadden, line 5: Nice :100: `let` allows us to reassign our variables

index.js
```diff
@@ -1,88 +1,121 @@
...
 
 // ***** Question 2 *****
-// *** Uncomment the lines below to test
+// *** Uncomment the lines below to teﬂst
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
 // console.log("%cQuestion 2", "color: red") 
 
-// console.log(name)
+const name = "Hanna"
```
seanpadden, line 17: Great! The key here was using `const` because it doesn't allow us to reassign our variable

index.js
```diff
@@ -1,88 +1,121 @@
...
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
 // console.log("%cQuestion 3", "color: red") 
-
-// console.log(drinkWater(12))
-// // => "Man I sure am thirsty"
-// // => "Ahh that hits the spot"
-// // => 11
-// console.log("%c----------", "color: red") 
+function drinkWater(current_thirst_level){
```
seanpadden, line 30: Well done. For Javascript, it's standard to use [camel case](https://upload.wikimedia.org/wikipedia/commons/thumb/c/c8/CamelCase_new.svg/1200px-CamelCase_new.svg.png) for function and variable names. So instead of `current_thirst_level`, name it `currentThirstLevel`

One way to condense this and not lose any readability would be to consolidate lines 32 and 33 to:
```js
return currentThirstLevel - 1
```

```javascript
```
```javascript
```

index.js
```diff
@@ -1,88 +1,121 @@
...
 
 
 // ***** Question 4 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 4", "color: red") 
 
-// console.log(sameSameButDifferent(5, 5)) 
-// // => "same same"
+console.log("%cQuestion 4", "color: red") 
+function sameSameButDifferent(num, maybeNum){
```
seanpadden, line 48: Nice! Another solution would be to convert the string to a number with parseInt to avoid `==` :

index.js
```diff
@@ -1,88 +1,121 @@
...
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 5", "color: red")
-// const student1 = { name: "Duane", grade: 88 }
-// updateGrade(student1, 92)
-// console.log(student1)
-// // => { name: "Duane", grade: 92 }
-// console.log("%c----------", "color: red")
+console.log("%cQuestion 5", "color: red")
+
+function updateGrade(student, grade){
```
seanpadden, line 78: Good :+1: The updates made to the `student` object inside the function is reflected in the `student1` object because objects are ["call by sharing" unlike primitive data types](https://stackoverflow.com/questions/518000/is-javascript-a-pass-by-reference-or-pass-by-value-language).

index.js
```diff
@@ -1,88 +1,121 @@
...
-//   },
-//   { 
-//     name: "Liza", phones: { cell: "555-234-5678", office: "555-567-1234" }
-//   }
-// ]
-// printNameAndPhones(users)
-// // => "Duane"
+console.log("%cQuestion 6", "color: red")
+
+function printNameAndPhones(user){
```
seanpadden, line 93: Nice. I would consider renaming the anonymous function's parameter on line 94 to something that's a little more indicative of what it's going to actually represent. `user_hash`, aside from using Ruby naming conventions (remember hashes are objects in Javascript!) is actually misleading because it's going to represent each element of our `users` array. So a better name might be `user`! 

You're still accomplishing the same thing and this is ultimately correct, but try to be as intentional as you can with naming.

index.js
```diff
@@ -141,29 +181,41 @@ const line = []
 
 
 // *** Uncomment the lines below to test
-// console.log("%cScope & Closures - Question 1", "color: red")
+console.log("%cScope & Closures - Question 1", "color: red")
 
-// console.log(takeATicketNumber(line))
-// // => `Welcome. You are ticket number 1`
+let ticketNo = 0
+function takeATicketNumber(line){
```
seanpadden, line 187: good! great use of global scope.

index.js
```diff
@@ -141,29 +181,41 @@ const line = []
...
+// => `Welcome. You are ticket number 3`
 
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red") 
 
 // ***** Scope & Closures - Question 2 *****
 // *** Uncomment the lines below to test
 // console.log("%cScope & Closures - Question 2", "color: red")
 
+// function ticketNumberGeneratorFunc(arr){
```
seanpadden, line 213: This one's a tricky bonus one! Check this Scrimba out https://scrimba.com/scrim/cLgEmZUZ and see if you can figure this out after watching it and playing around with the examples. (In Scrimba, you can pause the video and play around with the code shown on the screen)

PR #106, opened 2021-04-06
done - henry velasquez

HenPVel:


index.js
```diff
@@ -1,88 +1,125 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+var counter = 1
```
seanpadden, line 3: This works, but the point here is to demonstrate that we can use `let` to declare a variable and then we're able to reassign it. So you could do:

```javascript
let counter = 1
counter = 2
```

It may seem trivial in this example, but best to get in the habit as you'll see its use down the line!

index.js
```diff
@@ -1,88 +1,125 @@
...
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red") 
+
 
-// console.log(name)
+const name = 'Henry Velasquez'
```
seanpadden, line 21: Great! the key here was using `const` because it doesn't allow us to reassign our variable

index.js
```diff
@@ -1,88 +1,125 @@
...
 
 
 
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
 
-// console.log(drinkWater(12))
+function drinkWater(currentThirstLevel) {
```
seanpadden, line 34: Great 👍  One way to condense this and not lose any readability would be to consolidate lines 36 and 38 to:
```js
return currentThirstLevel - 1
```
- Also if we don't have the `return` and we try to use this function for getting a value, it will not work as expected as demonstrated in the following example:
```javascript
```
```javascript
```

index.js
```diff
@@ -1,88 +1,125 @@
...
 // ***** Question 4 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 4", "color: red") 
+console.log("%cQuestion 4", "color: red") 
+
+function sameSameButDifferent(num, maybeNum) {
+  if (num === maybeNum) {
+    console.log('same same')
+  } else if (num === parseInt(maybeNum, 10)) {
+    console.log( 'same same but different' ) //DO I PUT SPACES OR NO?
```
seanpadden, line 56: No spaces necessary! But works just the same.

index.js
```diff
@@ -1,88 +1,125 @@
...
+console.log("%c----------", "color: red") 
 
 
 
 // ***** Question 4 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 4", "color: red") 
+console.log("%cQuestion 4", "color: red") 
+
+function sameSameButDifferent(num, maybeNum) {
```
seanpadden, line 52: Nice! I also think it's worth exploring the difference between `===` and `==`. Another solution could look like:

```javascript
  if (num === maybeNum) {
    return "same same"
  }
  else if (num == maybeNum) {
  }
  else {
    return "different"
  }
}
```

index.js
```diff
@@ -1,88 +1,125 @@
...
 // ***** Question 5 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 5", "color: red")
-// const student1 = { name: "Duane", grade: 88 }
-// updateGrade(student1, 92)
-// console.log(student1)
+console.log("%cQuestion 5", "color: red")
+
+function updateGrade(student, grade) {
+  student['grade'] = grade   // WHAT EXACTLY IS THE CORRECT JSON OBJECT NOTATION. I TRIED [grade:] and [:grade] and they did not work
```
seanpadden, line 82: Good! 👍  

- **JSON** stands for **Javascript Object Notation** What you're doing here on line 82 is referred to as **bracket notation** (as opposed to **dot notation**) [you can see more here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_accessors)

You can also use dot notation like so:
`student.grade = grade`

The updates made to the `student` object inside the function is reflected in the `student1` object because objects are ["call by sharing" unlike primitive data types](https://stackoverflow.com/questions/518000/is-javascript-a-pass-by-reference-or-pass-by-value-language).

index.js
```diff
@@ -1,88 +1,125 @@
...
+
+function printNameAndPhones(users) {
+  users.forEach(function (e) {
+    console.log(e.name);
+    console.log(`Cell: ${e.phones['cell']}`);
+    console.log(`Office: ${e.phones['office']}`);
+  }
+  )
+}
+
```
seanpadden, line 105: Looks good! I would recommend using a more semantic name for the anonymous function's parameter on line 98 (like `user`) 

- They're equally good options, but here's how it would look with for..of:
```javascript
 for (let user of users) {
  console.log(user.name);
 }
```

- PS there's really no need to use bracket notation unless your key is a very long string or has spaces in it (like `['cell phone number']` but again, works just as good as dot notation.

index.js
```diff
@@ -141,44 +184,72 @@ const line = []
 
 
 // *** Uncomment the lines below to test
-// console.log("%cScope & Closures - Question 1", "color: red")
+console.log("%cScope & Closures - Question 1", "color: red")
 
-// console.log(takeATicketNumber(line))
+let ticket = 0
+
```
seanpadden, line 190: Good! nice use of global scope.

index.js
```diff
@@ -141,44 +184,72 @@ const line = []
...
+}
 
-// console.log(takeATicketNumberFunc(newLine))
-// // => `Welcome. You are ticket number 2`
+console.log("%cScope & Closures - Question 2", "color: red")
 
-// console.log(nowServing(newLine))
-// // => `Currently serving 1.`
+function ticketNumberGeneratorFunc() {
+  let ticket = 0
```
seanpadden, line 229: Excellent! Nice closure. 👍

PR #107, opened 2021-04-06
done

ajnikrant:
I hope this works! Submission attempt

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,80 +1,117 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+var counter = 1
+var counter = 2
+console.log(counter) 
+// => 2
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red") 
 
-// console.log(name)
+let name = "Amanda"
+console.log(name)
 // => "Raffy"
 // name = "Not Raffy" 
 // => TypeError
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red") 
 
 
 
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
 
-// console.log(drinkWater(12))
-// // => "Man I sure am thirsty"
-// // => "Ahh that hits the spot"
-// // => 11
-// console.log("%c----------", "color: red") 
+let drinkWater = function(num){
+  console.log("Man I sure am thirsty")
+  console.log("Ahh that hits the spot")
+  console.log(num-=1)
+}
+console.log(drinkWater(12))
+// => "Man I sure am thirsty"
+// => "Ahh that hits the spot"
+// => 11
+console.log("%c----------", "color: red") 
 
 
 
 // ***** Question 4 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 4", "color: red") 
+console.log("%cQuestion 4", "color: red") 
 
-// console.log(sameSameButDifferent(5, 5)) 
-// // => "same same"
+let sameSameButDifferent = function(num, maybeNum){
+  let result;
+  if (num === maybeNum){
+    result = console.log("same same");
+  } else if (num == maybeNum){
+    result = console.log("same same (but different)");
+  } else if (num !== maybeNum){
+    result = console.log("different");
+  }
+  return result;
+}
 
-// console.log(sameSameButDifferent(123, "123")) 
+
+console.log(sameSameButDifferent(5, 5)) 
+// => "same same"
+
+console.log(sameSameButDifferent(123, "123")) 
 // // => "same same (but different)"
 
-// console.log(sameSameButDifferent(5, 7)) 
+console.log(sameSameButDifferent(5, 7)) 
 // // => "different"
 
-// console.log(sameSameButDifferent(123, "122")) 
-// // => "different"
-// console.log("%c----------", "color: red") 
+console.log(sameSameButDifferent(123, "122")) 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,88 +1,125 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+var counter = 1
+var counter = 2
```
seanpadden, line 6: This does work, but the idea here is to get away from using `var` and instead use `let` so we're only reassigning the variables we mean to! So you could instead do

`let counter = 1`
`counter = 2`

This isn't a big deal in such a small example, but better to get into the habit now!

index.js
```diff
@@ -1,88 +1,125 @@
...
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red") 
 
-// console.log(name)
+let name = "Amanda"
```
seanpadden, line 18: On this one, we don't want to be able to reassign `name`! That's where `const` comes in. See what happens if you change `let` to `const` here. [You can read more on some differences here! ](https://medium.com/javascript-scene/javascript-es6-var-let-or-const-ba58b8dcde75#:~:text=%60const%60%20is%20a%20signal%20that,always%20the%20entire%20containing%20function.)

index.js
```diff
@@ -1,88 +1,125 @@
...
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
 
-// console.log(drinkWater(12))
-// // => "Man I sure am thirsty"
-// // => "Ahh that hits the spot"
-// // => 11
-// console.log("%c----------", "color: red") 
+let drinkWater = function(num){
```
seanpadden, line 31: - If we don't have the `return` and we try to use this function for getting a value, it will not work as expected as demonstrated in the following example:
```javascript
```
```javascript
```

index.js
```diff
@@ -1,88 +1,125 @@
...
+// => "Ahh that hits the spot"
+// => 11
+console.log("%c----------", "color: red") 
 
 
 
 // ***** Question 4 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 4", "color: red") 
+console.log("%cQuestion 4", "color: red")
```
seanpadden, line 46: Overall good! The key here is that you understand the difference between `===` and `==` which you're demonstrating. One flag I'll raise though is setting variables equal to a console log and then returning. 

While what you're doing does work, you're making extra work for yourself. You could refactor this to read: 

```javascript
  if (num === maybeNum) {
    return "same same"
  }
  else if (num == maybeNum) {
  }
  else {
    return "different"
  }
}
```
Remember for this deliverable, we just need to return the strings - not console log!

index.js
```diff
@@ -1,88 +1,125 @@
...
 // ***** Question 5 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 5", "color: red")
-// const student1 = { name: "Duane", grade: 88 }
-// updateGrade(student1, 92)
-// console.log(student1)
-// // => { name: "Duane", grade: 92 }
-// console.log("%c----------", "color: red")
+
+let updateGrade = function(student, grade1){
```
seanpadden, line 79: Great! You can also use dot notation like so:
`student.grade = grade`

and no need to set it equal to a new variable on line 80. 

The key here was understanding that objects are reference types. Object values are references, so when we mutate the student object inside updateGrade, the change will persist outside the function. Objects are ["call by sharing" unlike primitive data types](https://stackoverflow.com/questions/518000/is-javascript-a-pass-by-reference-or-pass-by-value-language)
```javascript
function updateGrade(studentObject, gradeNumber) {
 studentObject.grade = gradeNumber
}
```

index.js
```diff
@@ -1,88 +1,125 @@
...
-
-// const users = [ 
-//   { 
-//     name: "Duane", phones: { cell: "555-123-4567", office: "555-456-7890" }
-//   },
-//   { 
-//     name: "Liza", phones: { cell: "555-234-5678", office: "555-567-1234" }
-//   }
-// ]
+const printNameAndPhones = function(user){
```
seanpadden, line 95: Interesting implementation! This works, but you could simplify it by running something like:

```javascript
  console.log(user.name);
 })
```
- And here's how it'd look with a `for..of`
```javascript
 for (let user of users) {
  console.log(user.name);
 }
```

index.js
```diff
@@ -141,44 +186,60 @@ const line = []
 
 
 // *** Uncomment the lines below to test
-// console.log("%cScope & Closures - Question 1", "color: red")
 
-// console.log(takeATicketNumber(line))
-// // => `Welcome. You are ticket number 1`
+const takeATicketNumber = function(line){
```
seanpadden, line 190: Not seeing `line` defined anywhere? What datatype would `line` be?

index.js
```diff
@@ -141,44 +186,60 @@ const line = []
...
-// const takeATicketNumberFunc = ticketNumberGeneratorFunc()
-// console.log(takeATicketNumberFunc(newLine))
-// // => `Welcome. You are ticket number 1`
+const newLine = []
+
+// let ticketNumberGeneratorFunc = function(newLine){
+//   return `${newLine}`;
+// }
+
+const takeATicketNumberFunc = function ticketNumberGeneratorFunc(newLine){
```
seanpadden, line 225: This one's a tricky bonus one! Check this Scrimba out https://scrimba.com/scrim/cLgEmZUZ and see if you can figure this out after watching it and playing around with the examples. (In Scrimba, you can pause the video and play around with the code shown on the screen)

Also think about what `newLine` should be - does it need to be an array or could we use something else?

seanpadden:
Hi Amanda, 

See my comments on your code for specific feedback, but in a general sense I'd definitely like you to study up on the differences between `let` vs `const` a bit more. While it may not seem super important at the moment, you could definitely be more consistent and intentional in when you're using one or the other. I would also get into the habit of regular function syntax like:

```javascript
function myFunction(myArg) {
  return myArg
}
```

using `const` to define functions is fine, but I think it might get a little confusing for you. You're also almost never going to want to use `let` to define your functions because you could accidentally reassign them and then not have those functions anymore! It could cause a lot of headache & heartache for you down the line. Think of the `function` keyword as just another way we assign names to functions.

PR #108, opened 2021-04-06
Done?

majormatt28:
Done from last night mini challenge.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,44 +1,58 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
-
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+console.log("%cQuestion 1", "color: red") 
+var counter = 1;
+var counter = 2;
+console.log(counter) 
+ // => 2
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
-
-// console.log(name)
+console.log("%cQuestion 2", "color: red") 
+const name = "Raffy"
+console.log(name)
 // => "Raffy"
 // name = "Not Raffy" 
 // => TypeError
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red") 
 
 
 
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
 
+var phrase = (`Man I sure am thirsty`)
+var secondPhrase = ('Ahh that hits the spot')
+
+function drinkWater(currentThirstLevel) {
+  for (let i = 0; i < currentThirstLevel; --i){
+    return phrase = (`Man I sure am thirsty`)
+  }
+}
+console.log(drinkWater(12))
 // console.log(drinkWater(12))
-// // => "Man I sure am thirsty"
-// // => "Ahh that hits the spot"
-// // => 11
-// console.log("%c----------", "color: red") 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,44 +1,58 @@
...
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
-
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+console.log("%cQuestion 1", "color: red") 
+var counter = 1;
+var counter = 2;
+console.log(counter)
```
seanpadden, line 6: Remember we want to get away from using `var`! If you know you're going to reassign your variable, it's better to use `let`. Then you could do something like...

```javascript
let counter = 1
counter = 2
```

Best to get in the habit now and avoid using var!

index.js
```diff
@@ -1,44 +1,58 @@
...
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
-
-// console.log(name)
+console.log("%cQuestion 2", "color: red") 
+const name = "Raffy"
```

seanpadden, line 16: Great! The key here was using const because it doesn't allow us to reassign our variable

Michelle will give you bonus points for using Raffy.

index.js
```diff
@@ -1,44 +1,58 @@
...
 
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
 
+var phrase = (`Man I sure am thirsty`)
+var secondPhrase = ('Ahh that hits the spot')
+
+function drinkWater(currentThirstLevel) {
```
seanpadden, line 32: This approach would work if our `currentThirsLevel` was an array that we wanted to iterate over and return something for each element of the array, but that's not quite what we're going for here:

```javascript
function drinkWater(currentThirstLevel) {
  console.log("Man, I sure am thirsty")
  currentThirstLevel -= 1
  return currentThirstLevel
}
```

index.js
```diff
@@ -1,44 +1,58 @@
...
 
 
 // ***** Question 4 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 4", "color: red") 
+console.log("%cQuestion 4", "color: red") 
 
-// console.log(sameSameButDifferent(5, 5)) 
-// // => "same same"
+function sameSameButDifferent(num, maybeNum){
```
seanpadden, line 50: Think about the `===` and `==` operators - how could we use both of these to achieve the deliverable here?

index.js
```diff
@@ -48,18 +62,23 @@
...
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 5", "color: red")
-// const student1 = { name: "Duane", grade: 88 }
-// updateGrade(student1, 92)
-// console.log(student1)
+console.log("%cQuestion 5", "color: red")
+const student1 = { name: "Duane", grade: 88 }
+updateGrade(student1, 92)
+
+function updateGrade(student, grade) {
```
seanpadden, line 75: in Ruby, how would we update a value on a hash? Maybe something like...

`student[:grade] = 12` 

How about in Javascript? [Check out property accessors for one or two ways to do this!](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_accessors)

seanpadden:
Hey Matt,

See my comments for more specific feedback on the code you submitted, but make sure to go back and study iterators ([forEach](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) and/or for loops especially), the differences between let vs. const, our [equality operators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness), and [scope](https://developer.mozilla.org/en-US/docs/Glossary/Scope).

The scrimba videos that are in the syllabus from Day One should help reinforce some of these concepts!

PR #109, opened 2021-04-06
Done

coltenappleby:


index.js
```diff
@@ -1,88 +1,134 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+var counter = 1
+var counter = 2
```
seanpadden, line 6: This works, but we should get into the practice of using `let` when we want to reassign a variable. It will save you a lot of headache and heartache down the line! You could do something like..

```javascript
let counter = 1
counter = 2
```

index.js
```diff
@@ -1,88 +1,134 @@
...
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red") 
+
+const name = "Colten"
```
seanpadden, line 19: Good! Using `const` so we can't reassign 👍

index.js
```diff
@@ -1,88 +1,134 @@
...
 
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
+
+
+
+function drinkWater(current_thirt_level) {
+  var thirst = current_thirt_level
```
seanpadden, line 36: Well done. For Javascript, it's standard to use [camel case](https://upload.wikimedia.org/wikipedia/commons/thumb/c/c8/CamelCase_new.svg/1200px-CamelCase_new.svg.png) for function and variable names. So instead of `current_thirst_level`, name it `currentThirstLevel`

- No need to create a new variable `thirst` since we're really only using it to change the value of our `currentThirsLevel`!

- One way to condense this and not lose any readability would be to consolidate lines 38 and 40 to:
```js
return currentThirstLevel - 1
```

index.js
```diff
@@ -1,88 +1,134 @@
...
 
 
 // ***** Question 4 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 4", "color: red") 
+console.log("%cQuestion 4", "color: red") 
 
-// console.log(sameSameButDifferent(5, 5)) 
-// // => "same same"
+function sameSameButDifferent(num, maybeNum) {
```
seanpadden, line 57: Nice. The key here was understanding the difference between `===` and `==`. You could also convert the string to a number with parseInt to avoid == :

index.js
```diff
@@ -1,88 +1,134 @@
...
 // ***** Question 5 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 5", "color: red")
-// const student1 = { name: "Duane", grade: 88 }
-// updateGrade(student1, 92)
-// console.log(student1)
-// // => { name: "Duane", grade: 92 }
-// console.log("%c----------", "color: red")
+
+function updateGrade(student_object, grade) {
```
seanpadden, line 88: Good :+1: The updates made to the `student` object inside the function is reflected in the `student1` object because objects are ["call by sharing" unlike primitive data types](https://stackoverflow.com/questions/518000/is-javascript-a-pass-by-reference-or-pass-by-value-language).

index.js
```diff
@@ -1,88 +1,134 @@
...
-//     name: "Duane", phones: { cell: "555-123-4567", office: "555-456-7890" }
-//   },
-//   { 
-//     name: "Liza", phones: { cell: "555-234-5678", office: "555-567-1234" }
-//   }
-// ]
-// printNameAndPhones(users)
+console.log("%cQuestion 6", "color: red")
+
+function printNameAndPhones(users) {
```
seanpadden, line 107: This for loop works, but for iterating over arrays, Javascript has a couple other methods that make it a little easier.
```javascript
 for (let user of users) {
  console.log(user.name);
 }
```
```javascript
  console.log(user.name);
 })
```

index.js
```diff
@@ -100,23 +146,31 @@ function myMap(array, callback) {
...
 // ***** Callbacks - Question 2 *****
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 2", "color: red")
+console.log("%cCallbacks - Question 2", "color: red")
 
-// console.log(myMap(["Raffy", "Chase"], greet)) 
-// // => ["Hi, Raffy!", "Hi, Chase!"]
-// console.log("%c----------", "color: red") 
+function greet (name) {
+  return "Hi, " + name + "!"
```
seanpadden, line 168: cool! You can also interpolate with `${name}`

index.js
```diff
@@ -138,27 +192,33 @@ function nowServing(line) {
 const line = []
 
 // ***** Scope & Closures - Question 1 *****
+var line_length = 0
+function takeATicketNumber(line) {
```
seanpadden, line 196: nice - great use of global scope!

index.js
```diff
@@ -138,27 +192,33 @@ function nowServing(line) {
...
 
-// console.log(takeATicketNumber(line))
-// // => `Welcome. You are ticket number 3`
+console.log(takeATicketNumber(line))
+// => `Welcome. You are ticket number 3`
 
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red") 
 
 // ***** Scope & Closures - Question 2 *****
```
seanpadden, line 223: This one's a tricky bonus one! Check this Scrimba out https://scrimba.com/scrim/cLgEmZUZ and see if you can figure this out after watching it and playing around with the examples. (In Scrimba, you can pause the video and play around with the code shown on the screen)

seanpadden:
Nice work Colten. See my comments for some specific feedback on your code but a few things to summarize: 

- using `let` and `const` properly 
- getting into the habit of Javascript's camel case
- getting comfortable with callbacks and iterators like `forEach`

PR #110, opened 2021-04-06
Stuck -I didn't finish

rsabur:


index.js
```diff
@@ -1,71 +1,106 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+let counter = 1
+counter = 2
+
```
BrendaMichelle, line 6: Nice :100: `let` allows us to reassign our variables

index.js
```diff
@@ -1,71 +1,106 @@
...
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red") 
 
-// console.log(name)
+const name = 'Radiyah'
+
```
BrendaMichelle, line 21: Great! The key here was using `const` because it doesn't allow us to reassign our variable

index.js
```diff
@@ -1,71 +1,106 @@
...
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
 
-// console.log(drinkWater(12))
-// // => "Man I sure am thirsty"
-// // => "Ahh that hits the spot"
-// // => 11
-// console.log("%c----------", "color: red") 
+function drinkWater(num) {
```
BrendaMichelle, line 34: Good work using [camel case](https://upload.wikimedia.org/wikipedia/commons/thumb/c/c8/CamelCase_new.svg/1200px-CamelCase_new.svg.png) for the function name.

index.js
```diff
@@ -1,71 +1,106 @@
...
+   console.log("Man I sure am thirsty");
+   newNum = num -= 1 ;
+   console.log("Ahh that hits the spot");
+   console.log(newNum)
+}
```
BrendaMichelle, line 39: Although this works, it's not good practice to assign variables like this. If you want to create a new variable, I would suggest:
```js
  let newNum = num -1;
```

In this case, you don't necessarily need a new variable though because you already have the variable `num` that you can re-assign.
```suggestion
   num -= 1 ;
   console.log("Ahh that hits the spot");
   return num
```

index.js
```diff
@@ -1,71 +1,106 @@
...
-// // => "same same"
+function sameSameButDifferent(num, maybeNum) {
+  if (num === maybeNum) {
+    return "same same"
+  } else if (num == maybeNum) {
+    return "same same (but different)"
+  } else {
+    return "different"
+  }
+}
```
BrendaMichelle, line 60: Nice! Another solution would be to convert the string to a number with parseInt to avoid `==` :

index.js
```diff
@@ -1,71 +1,106 @@
...
 
 // ***** Question 6 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 6", "color: red")
+console.log("%cQuestion 6", "color: red")
+
+// function printNameAndPhones(users) {
+//   users.forEach();
+// }
+
```
BrendaMichelle, line 103: Looks like you're on the right track! The `forEach` method should take a function as an argument. The function that it takes in as an argument should do what you would do to one user object. If you get stuck using `forEach`, consider also using a for loop or `for...of`

```javascript
 for (let user of users) {
  console.log(user.name);
 }
```

```javascript
  console.log(user.name);
 })
```

index.js
```diff
@@ -100,23 +135,33 @@ function myMap(array, callback) {
...
-// console.log("%cCallbacks - Question 2", "color: red")
+console.log("%cCallbacks - Question 2", "color: red")
+
 
-// console.log(myMap(["Raffy", "Chase"], greet)) 
-// // => ["Hi, Raffy!", "Hi, Chase!"]
-// console.log("%c----------", "color: red") 
+function greet(name) {
+  return `Hi, ${name}!`
+}
```
BrendaMichelle, line 159: Good! So these are callback functions when we're passing them in as an argument to the `myMap` function.

index.js
```diff
@@ -158,7 +209,7 @@ const line = []
 // console.log(takeATicketNumber(line))
 // // => `Welcome. You are ticket number 3`
 
-// console.log("%c----------", "color: red") 
+console.log("%c----------", "color: red") 
 
 // ***** Scope & Closures - Question 2 *****
 // *** Uncomment the lines below to test
```
BrendaMichelle, line 215: On the right track! Re-attempt these last two now that we've had the review :)

BrendaMichelle (commented):
Left some comments

PR #111, opened 2021-04-06
almost done, will keep working on bonus' later

susanavik:


index.js
```diff
@@ -13,18 +14,32 @@
 // *** otherwise, the error will stop the rest of your code from running
 // console.log("%cQuestion 2", "color: red") 
 
-// console.log(name)
+const name = "Raffy"
```
BrendaMichelle, line 17: Good! The key here was using `const` because it doesn't allow us to reassign our variable

index.js
```diff
@@ -2,7 +2,8 @@
 // *** Uncomment the lines below to test
 // console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
+let counter = 1
```
BrendaMichelle, line 5: Good! Declaring the variable with `let` means that we can reassign the variable later

index.js
```diff
@@ -13,18 +14,32 @@
...
 // => TypeError
 // console.log("%c----------", "color: red") 
 
 
-
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
 // console.log("%cQuestion 3", "color: red") 
 
+let currentThirstLevel = 12
```
BrendaMichelle, line 29: No need for this variable. When invoking `drinkWater` on line 43, the number `12` is being passed in as an argument, which is then assigned as the value to the parameter `currentThirstLevel` on line 31

index.js
```diff
@@ -13,18 +14,32 @@
...
+
+function drinkWater(currentThirstLevel) {
+  let currentThirstLevel = 12;
```
BrendaMichelle, line 32: same comment as the previous for line 12

index.js
```diff
@@ -13,18 +14,32 @@
...
 // *** Uncomment the lines below to test
 // console.log("%cQuestion 3", "color: red") 
 
+let currentThirstLevel = 12
+
+function drinkWater(currentThirstLevel) {
+  let currentThirstLevel = 12; 
+  while (currentThirstLevel > 11) 
+     return "Man I sure am thirsty";
+     currentThirstLevel -= 1;
```
BrendaMichelle, line 35: - There's no opening curly brace to open the block for the while loop. That will cause an error. Be sure to test your code since the curly braces in the code will not match up
- I want to point out that any code written after a `return` statement will not run. When JavaScript sees the `return` it exists the function
- However, you don't need a while loop for this solution (The Ruby method you're meant to translate from doesn't have a while loop).

index.js
```diff
@@ -13,18 +14,32 @@
...
 // console.log("%cQuestion 3", "color: red") 
 
+let currentThirstLevel = 12
+
+function drinkWater(currentThirstLevel) {
+  let currentThirstLevel = 12; 
+  while (currentThirstLevel > 11) 
+     return "Man I sure am thirsty";
+     currentThirstLevel -= 1; 
+  } else if (currentThirstLevel = 11) {
```
BrendaMichelle, line 36: No need for conditionals in this solution either.

index.js
```diff
@@ -37,6 +52,17 @@
...
+  if (num === maybeNum) {
+    return 'same same'
+  } else if { num == maybeNum) {
+    return 'same same (but different)'
+  } else {
+    return "different"
+  }
+    
+}
+
```
BrendaMichelle, line 65: Nice! Another solution would be to convert the string to a number with parseInt to avoid `==` :

index.js
```diff
@@ -62,11 +88,24 @@
 // console.log("%c----------", "color: red")
 
 
+function updateGrade (student1, grade) {
+  student1.grade = grade;
+
+  return student1;
+}
```
BrendaMichelle, line 95: Good!

index.js
```diff
@@ -62,11 +88,24 @@
...
 // *** Uncomment the lines below to test
 // console.log("%cQuestion 6", "color: red")
 
+function printNameandPhones(users) {
+  users.forEach(function (userObject) {
+    console.log(userObject.name);
+    console.log('Cell: ${userObject.phones.cell}')
+    console.log('Office: ${userObject.phones.office}')
+  })
+}
```
BrendaMichelle, line 107: Very nice! It's ok to use semicolons or to not use them, just be consistent

index.js
```diff
@@ -114,7 +157,11 @@ function myMap(array, callback) {
 // *** Uncomment the lines below to test
 // console.log("%cCallbacks - Question 2", "color: red")
 
-// console.log(myMap(["Raffy", "Chase"], greet)) 
+function greet(nameStr) {
+  return 'Hi, ${nameSt)!'
+}
+
```
BrendaMichelle, line 163: 💯

index.js
```diff
@@ -138,7 +185,14 @@ function nowServing(line) {
...
 
 // ***** Scope & Closures - Question 1 *****
+let ticket = 0
 
+function takeATicketNumber(line) {
+  ticket +=1
+  line.push(ticket) 
+
+  returrn 'Welcome. You are ticket number ${ticket}'
+}
```
BrendaMichelle, line 195: Good! The key here was declaring the ticket variable in the global scope

BrendaMichelle (commented):
Left some comments

PR #112, opened 2021-04-06
Done

jabair14:


index.js
```diff
@@ -1,88 +1,123 @@
 // ***** Question 1 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
+// let counter = 1;
+const counter = 2
```
BrendaMichelle, line 6: For this, you should use `let` instead instead of `const` because `let` allows us to reassign our variables

index.js
```diff
@@ -1,88 +1,123 @@
...
+console.log("%c----------", "color: red") 
 
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+const name = "Jabair"
+console.log("%cQuestion 2", "color: red")
```
BrendaMichelle, line 21: Great! The key here was using `const` because it doesn't allow us to reassign our variable

index.js
```diff
@@ -1,88 +1,123 @@
...
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
+  function drinkWater(current_thirst_level){
+    console.log("Man I sure am thirsty")
+    current_thirst_level -= 1
+    console.log("Ahh that hits the spot")
+    return current_thirst_level
+  }
```
BrendaMichelle, line 39: Well done. For Javascript, it's standard to use [camel case](https://upload.wikimedia.org/wikipedia/commons/thumb/c/c8/CamelCase_new.svg/1200px-CamelCase_new.svg.png) for function and variable names. So instead of `current_thirst_level`, name it `currentThirstLevel`

index.js
```diff
@@ -1,88 +1,123 @@
...
+
+function sameSameButDifferent(num, maybeNum){
+  if (num === maybeNum){
+    return "same same"
+  } else if (num == maybeNum) {
+    return "same same (but different)"
+  } else {
+    return "different"
+  }
+}
```

index.js
```diff
@@ -1,88 +1,123 @@
...
+console.log("%cQuestion 5", "color: red")
+const student1 = { name: "Duane", grade: 88 }
+
+function updateGrade (student, grade) {
+  student.grade = grade 
+}
+updateGrade(student1, 92)
+console.log(student1)
+// => { name: "Duane", grade: 92 }
+console.log("%c----------", "color: red")
```
BrendaMichelle, line 89: Good!

index.js
```diff
@@ -1,88 +1,123 @@
...
+]
+printNameAndPhones(users)
+
+function printNameAndPhones(users){
+  users.forEach(function(userInfo){
+    console.log(userInfo.name);
+    console.log(userInfo.phones.cell);
+    console.log(userInfo.phones.office)
+  })
+}
```
BrendaMichelle, line 113: Good! If you use semicolons, be consistent 🙂

index.js
```diff
@@ -100,10 +135,13 @@ function myMap(array, callback) {
...
+newArr = myMap
+
 
-// console.log(myMap([1,2,3,4,5], triple)) 
-// // => [3,6,9,12,15]
+console.log(myMap([1,2,3,4,5], triple)) 
+// => [3,6,9,12,15]
 
 // console.log(myMap([2,4,6,8], triple))   
 // // => [6,12,18,24]
```
BrendaMichelle, line 147: Where you able to figure these out after the review?

BrendaMichelle (commented):
Left some comments

index.js
jabair14: Yes it made more sense after the review, i think i just need to spend more time with the callbacks.

PR #113, opened 2021-04-06
Partial Done

yuringn:


index.js
```diff
@@ -1,88 +1,126 @@
...
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
-
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+console.log("%cQuestion 1", "color: red") 
+let counter = 1;
+counter = 2;
+console.log(counter)
```
BrendaMichelle, line 9: Nice :100: `let` allows us to reassign our variables

index.js
```diff
@@ -1,88 +1,126 @@
...
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
-
-// console.log(name)
+console.log("%cQuestion 2", "color: red") 
+const name = "Yuri";
```

index.js
```diff
@@ -1,88 +1,126 @@
...
 
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 3", "color: red") 
+console.log("%cQuestion 3", "color: red") 
+function drinkWater(currentThirstLevel){
+  console.log ("Man I sure am thirsty")
+  currentThirstLevel -=1
+  return "Ahh that hits the spot"
+}
```
BrendaMichelle, line 35: Good. Way to use  [camel case](https://upload.wikimedia.org/wikipedia/commons/thumb/c/c8/CamelCase_new.svg/1200px-CamelCase_new.svg.png) for your variable and function names 👍

index.js
```diff
@@ -1,88 +1,126 @@
...
+function sameSameButDifferent(num, maybeNum){
+  if (num === maybeNum){
+    return "same same"
+  }
+  if (num == maybeNum){
+  return "same same (but different)"
+}else{
+  return "different"
+}
+}
```
BrendaMichelle, line 58: Nice! Another solution would be to convert the string to a number with parseInt to avoid `==` :

index.js
BrendaMichelle: For this solution, the deliverable is expecting the return value to be a number instead of a string.

index.js
```diff
@@ -1,88 +1,126 @@
...
-// const student1 = { name: "Duane", grade: 88 }
-// updateGrade(student1, 92)
-// console.log(student1)
+console.log("%cQuestion 5", "color: red")
+const student1 = { name: "Duane", grade: 88 }
+function updateGrade(studen,grade){
+//  return Object.assign(student,grade)
+
+}
+updateGrade(student1, 92)
```
BrendaMichelle, line 82:
```suggestion
const student1 = { name: "Duane", grade: 88 }
function updateGrade(student, grade){
  student.grade = grade

}
updateGrade(student1, 92)
```


`student.grade` inside the function - this is accessing the grade property within the student object. Then ` = grade` is re-assigning the value of that property to the value of `grade` that we receive as an argument to the function.

index.js
```diff
@@ -1,88 +1,126 @@
...
+// }
+// )
+
+function printNameAndPhones(details){
+  console.log(details.name)
+  console.log("Cell:", details.phones.cell)
+  console.log("Office:", details.phones.office)
+}
+
+users.forEach(printNameAndPhones)
```
BrendaMichelle, line 114: This works! However, the deliverable was looking for a function that took in an array of users as an argument. Within that function you would iterate over the `users` array console logging each user's information.

index.js
```diff
@@ -100,14 +138,16 @@ function myMap(array, callback) {
...
 
 // *** Uncomment the lines below to test
-// console.log("%cCallbacks - Question 1", "color: red")
-
-// console.log(myMap([1,2,3,4,5], triple)) 
+console.log("%cCallbacks - Question 1", "color: red")
+function triple(myMap) {
+  console.log(myMap * 3)
+}
+
```
BrendaMichelle, line 145: This looks like it'll work. However, this function is being passed in a number (not a callback). This function should solely take in a number and return it tripled.

BrendaMichelle (commented):
Keep up the good work! Left some comments. Were you able to finish the remaining ones after the review?

PR #116, opened 2021-04-06
Done

amautacelt:


index.js
```diff
@@ -1,65 +1,98 @@
...
 // *** Uncomment the lines below to test
-// console.log("%cQuestion 1", "color: red") 
+console.log("%cQuestion 1", "color: red") 
 
-// console.log(counter) 
-// // => 2
-// console.log("%c----------", "color: red") 
+let counter = 1
+counter = 2
+console.log(counter)
```
BrendaMichelle, line 7: Yes :100: `let` allows us to reassign our variables

index.js
```diff
@@ -1,65 +1,98 @@
...
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
 // *** after testing, comment the line causing the error back in 
 // *** otherwise, the error will stop the rest of your code from running
-// console.log("%cQuestion 2", "color: red") 
+console.log("%cQuestion 2", "color: red") 
 
-// console.log(name)
+const name = 'carlos'
+console.log(name)
```
BrendaMichelle, line 19: Good. The key here was using `const` because it doesn't allow us to reassign our variable

index.js
```diff
@@ -1,65 +1,98 @@
...
+// end
+
+function drinkWater(current_thirst_level) {
+  console.log("Man I sure am thirsty")
+  current_thirst_level -= 1
+  console.log("Ahh that hits the spot")
+  return current_thirst_level
+}
+
+console.log(drinkWater(12))
```
BrendaMichelle, line 45: Well done. For Javascript, it's standard to use [camel case](https://upload.wikimedia.org/wikipedia/commons/thumb/c/c8/CamelCase_new.svg/1200px-CamelCase_new.svg.png) for function and variable names. So instead of `current_thirst_level`, name it `currentThirstLevel`

index.js
```diff
@@ -1,65 +1,98 @@
...
+    return 'same same'
+  } 
+  else if (num == maybeNum) {
+    return 'same same (but different)'
+  }
+  else {
+    return 'different'
+  }
+}
+console.log(sameSameButDifferent(5, 5))
```
BrendaMichelle, line 69: Nice! Another solution would be to convert the string to a number with parseInt to avoid `==` :

BrendaMichelle (commented):
Good work! Were you able to get the remaining deliverables after the review? Message me on slack/message the questions channel if there's deliverables you need more clarity on

PR #129, opened 2021-07-26
progress

tjhan14:
Galaxy!

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,6 +6,9 @@
 // // => 2
 // console.log("%c----------", "color: red") 
 
+counter = 1
+counter = 2
+console.log(counter)
 
 // ***** Question 2 *****
 // *** Uncomment the lines below to test
@@ -19,7 +22,8 @@
 // => TypeError
 // console.log("%c----------", "color: red") 
 
-
+const name = "Bob" 
+console.log(name)
 
 // ***** Question 3 *****
 // *** Uncomment the lines below to test
@@ -32,6 +36,13 @@
 // console.log("%c----------", "color: red") 
 
 
+function drinkWater(current_thirst_level){
+  console.log("Man I sure am thirsty")
+  current_thirst_level -= 1
+  console.log("Ahh that hits the spot")
+  return current_thirst_level
+}
+console.log(drinkWater(12))
 
 // ***** Question 4 *****
 // *** Uncomment the lines below to test
@@ -50,7 +61,18 @@
 // // => "different"
 // console.log("%c----------", "color: red") 
 
-
+function sameSameButDifferent(num, maybeNum) {
+  if (num === maybeNum) {
+    return "same same"
+  }
+  else if (num == maybeNum) {
+    return "same same (but different)"
+  }
+  else {
+    return "different"
+  }
+}
+console.log(sameSameButDifferent(4, 3))
 
 // ***** Question 5 *****
 // *** Uncomment the lines below to test
@@ -61,7 +83,15 @@
 // // => { name: "Duane", grade: 92 }
 // console.log("%c----------", "color: red")
 
-
+function updateGrade(student, grade) {
+  student.grade = grade
+  }
+  
+  const student1 = { name: "Duane", grade: 88 }
+  
+  updateGrade(student1, 92)
+  
+  console.log(student1)  
 
 // ***** Question 6 *****
 // *** Uncomment the lines below to test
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -6,6 +6,9 @@
 // // => 2
 // console.log("%c----------", "color: red") 
 
+counter = 1
+counter = 2
```
BrendaMichelle, line 10: use `let` here to declare the variable. `let` is a keyword that we can use to tell JavaScript explicitly that we are creating a variable. `let` will give the variable "block" scope meaning that the variable will only exist within the two curly braces in which it's defined. `let` allows us to re-assign the variable as well.

```suggestion
let counter = 1
counter = 2
```

index.js
```diff
@@ -19,7 +22,8 @@
 // => TypeError
 // console.log("%c----------", "color: red") 
 
-
+const name = "Bob" 
+console.log(name)
```
BrendaMichelle, line 26: Good! `const` is another one of those keywords that we use to tell JavaScript that we are creating a variable. Like `let`, it's also block scoped. The special thing about `const` is that it does not allow us to re-assign the variable.

index.js
```diff
@@ -32,6 +36,13 @@
 // console.log("%c----------", "color: red") 
 
 
+function drinkWater(current_thirst_level){
+  console.log("Man I sure am thirsty")
+  current_thirst_level -= 1
+  console.log("Ahh that hits the spot")
+  return current_thirst_level
+}
```
BrendaMichelle, line 44: Very good! Good job using camelCase for the function name. It's also convention to use camelCase instead of snake_case for variable names in JavaScript.
```suggestion
function drinkWater(currentThirstLevel){
  console.log("Man I sure am thirsty")
  currentThirstLevel -= 1
  console.log("Ahh that hits the spot")
  return currentThirstLevel
}
```

index.js
```diff
@@ -50,7 +61,18 @@
...
+  if (num === maybeNum) {
+    return "same same"
+  }
+  else if (num == maybeNum) {
+    return "same same (but different)"
+  }
+  else {
+    return "different"
+  }
+}
```
BrendaMichelle, line 74: Nice! Another solution would be to convert the string to a number with parseInt to avoid using `==` in the second conditional statement :

index.js
```diff
@@ -61,7 +83,15 @@
...
-
+function updateGrade(student, grade) {
+  student.grade = grade
+  }
+  
+  const student1 = { name: "Duane", grade: 88 }
+  
+  updateGrade(student1, 92)
+  
+  console.log(student1)
```
BrendaMichelle, line 94: Good!

index.js
```diff
@@ -84,6 +114,34 @@
...
+  users.forEach(beef)
+}
+
+function beef(user){
+  console.log(user.name)
+  console.log(user.phones.cell)
+  console.log(user.phones.office)
+}
+
+print_name_and_phones(users)
```
BrendaMichelle, line 144: Nice job! Beef 😂

index.js
```diff
@@ -109,6 +167,23 @@ function myMap(array, callback) {
...
+    const element = array[i]
+    const value = callback(element)
+    result.push(value)
+  }
+  return result
+}
+
+function triple(num) {
+  return (num * 3)
+}
```
BrendaMichelle, line 182: Good! You don't need the parenthesis around the return value. However, with or without the parenthesis, it's good! (: 

```suggestion
function triple(num) {
  return num * 3
}
```

index.js
```diff
@@ -118,7 +193,8 @@ function myMap(array, callback) {
 // // => ["Hi, Raffy!", "Hi, Chase!"]
 // console.log("%c----------", "color: red") 
 
-
+function greet(name) {
+  return `Hi, ${name}!`
```
BrendaMichelle, line 198: also, good! But where does the function end? (: You need a closing parenthesis to end the function body

```suggestion
function greet(name) {
  return `Hi, ${name}!`
}
```

index.js
```diff
@@ -118,7 +193,8 @@ function myMap(array, callback) {
...
 
 // ***** Scope & Closures *****
```
BrendaMichelle, line 200: - For the first scope question, the key is to use a global variable. So try declaring `ticketNumber` outside of the function and updating it appropriately inside of the function 🤔

- The scope ones are tricky. Especially the bonus one! In order to help understand the the bonus one, check this Scrimba out https://scrimba.com/scrim/cLgEmZUZ and see if you can figure it out after watching it and playing around with the examples that Ian gives. (In Scrimba, you can pause the video and play around with the code shown on the screen. Just pause the video and click into the screen and start typing or running the code. Then, press play when you're ready to continue on.)

